PR #8, opened 2023-02-28
[7] data layer models

steve1rm:
# Pull Request template

###  What does this PR solve?
Creates data layer


data/src/main/java/me/androidbox/data/remote/model/AttendeeModel.kt
```diff
@@ -0,0 +1,10 @@
+package me.androidbox.data.remote.model
+
+data class AttendeeModel(
+    val email: String,
+    val fullName: String,
+    val userId: String,
+    val eventId: String,
+    val isGoing: Boolean,
+    val remindAt: Long
+)
```
philipplackner: If you follow the presentation/domain/data structure, I'd recommend to reserve model as a name for the domain layer. This model you have here represents a data model which is parsed from JSON in this case. My personal recommendation is to name data models from an API ending with Dto, so for example `AttendeeDto` (data transfer object), data models from a local DB Entity, so for example `AttendeeEntity` and models in your domain layer without and suffix, so just `Attendee`.
This is not a strict rule, but avoids any confusion with the names.

Same ofc for all other models in this PR

data/src/main/java/me/androidbox/data/remote/model/response/FullAgendaModel.kt
```diff
@@ -0,0 +1,7 @@
+package me.androidbox.data.remote.model.response
+
+data class FullAgendaModel(
+    val listOfEventId: List<String>,
+    val listOfTaskId: List<String>,
+    val listOfReminderId: List<String>
+)
```
philipplackner: The names of the fields need to match the names of the JSON, otherwise you'll get a crash

philipplackner (approved):

data/src/main/java/me/androidbox/data/remote/model/AttendeeModel.kt
steve1rm: @philipplackner 
Just quick question. So we should separate our data models for roomdb and api?
for example:
```
data layer (AttendeeDto, AttendeeEntity) -> for the API and DB
domain layer (AttendeeModel)
presentation (Atteendee)
```

MKiperszmidSF: Exactly
Tho the one from Presentation, is completely optional. Usually on Presentation you'll also use the one from Domain

data/src/main/java/me/androidbox/data/remote/model/response/FullAgendaDto.kt
```diff
@@ -0,0 +1,17 @@
...
+ * call `/agenda` instead
+ *
+ * */
+data class FullAgendaDto(
+    val deletedEventIds: List<String>,
+    val deletedTaskIds: List<String>,
+    val deletedReminderIds: List<String>
+)
+
+/** TODO This seems to return the same results as the fullAgenda */
```
steve1rm, line 17: Looking at the documentation this seems to be the same as the body in the /syncAgenda that is a POST request.

But /FullAgenda seems to send the same in the body which is a GET request.

Is this correct.

data/src/main/java/me/androidbox/data/remote/model/request/EventRequestDto.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Response body: EventDto
+ * */
+data class EventRequestDto(
+    val id: String,
+    val title: String,
+    val description: String,
+    val from: Long,
+    val to: Long,
+    val remindAt: Long,
+    val attendeeIds: List<String>
```
MKiperszmidSF, line 20: I recommend adding SerializedName annotation if using GSON, or  @field:Json if using Moshi

If not, if you end up changing the variable name, or the app gets encrypted (you release it to playstore), then this Dto won't work

data/src/main/java/me/androidbox/data/remote/model/request/EventRequestDto.kt
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * /event
+ * POST/Multipart
+ *
+ * Creates a new event. Since Tasky supports offline mode, the IDs for agenda items
+ * need to be generated client-side. Please use UUID.randomUUID().toString() for that.
+ *
+ * Response body: EventDto
+ * */
+data class EventRequestDto(
```
MKiperszmidSF, line 13: I recommend adding SerializedName annotation if using GSON, or  @field:Json if using Moshi

If not, if you end up changing the variable name, or the app gets encrypted (you release it to playstore), then this Dto won't work

This is to all your DTO's. I'll just leave 1 comment as to not disturb the whole pr

MKiperszmidSF (approved):
Left a recommendation for dto

data/src/main/java/me/androidbox/data/remote/model/request/EventRequestDto.kt
steve1rm: @MKiperszmidSF 
Is it enough just to annotate all the data classes with the following?

`@JsonClass(generateAdapter = true)`

As the field names are the same as what is expected from the JSON is it good practice to still do this to every field in all the data classes?

philipplackner: If the field names are the same, you don't need to add any annotations. I personally only use annotations if the name from the JSON field would be better understandable with a different name

philipplackner (approved):

PR #9, opened 2023-02-28
[8] data layer roomdb

steve1rm:
# Pull Request template

###  What does this PR solve?
Adds room db


data/src/main/java/me/androidbox/data/local/dao/AgendaDao.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import androidx.room.*
+import kotlinx.coroutines.flow.Flow
+import me.androidbox.data.local.DatabaseConstant
+import me.androidbox.data.remote.model.AgendaModel
+
+@Dao
+interface AgendaDao {
+    @Query("SELECT * FROM ${DatabaseConstant.AGENDA_TABLE}")
+    fun getAgenda(): Flow<List<AgendaModel>>
```
philipplackner: You never need to retrieve the whole agenda in this app, only for a specific day

data/src/main/java/me/androidbox/data/local/dao/AttendeeDao.kt
```diff
@@ -0,0 +1,21 @@
...
+import me.androidbox.data.local.DatabaseConstant
+import me.androidbox.data.remote.model.AttendeeModel
+
+@Dao
+interface AttendeeDao {
+    @Query("SELECT * FROM ${DatabaseConstant.ATTENDEE_TABLE}")
+    fun getAttendee(): Flow<List<AttendeeModel>>
+
+    @Insert(onConflict = OnConflictStrategy.IGNORE)
+    fun insertAttendee(attendeeModel: AttendeeModel)
```
philipplackner: Any reason for the `IGNORE` strategy? Wouldn't you rather want to replace them if they exist? That way you also wouldn't need the `@Update` function

data/src/main/java/me/androidbox/data/remote/model/request/EventRequestModel.kt
```diff
@@ -0,0 +1,11 @@
+package me.androidbox.data.remote.model.request
+
+data class EventRequestModel(
+    val id: String,
+    val title: String,
+    val description: String,
+    val from: Long,
+    val to: Long,
+    val remindAt: Long,
+    val listOfAttendee: List<String>
```
philipplackner: Personal opinion: Naming this just `attendees` says it all while being shorter. Same for your other list namings.
Also, if I remember correctly, an attendee is more than just a string here

philipplackner (approved):

data/src/main/java/me/androidbox/data/local/dao/AttendeeDao.kt
steve1rm: @philipplackner 
Just wondering, if we use replace instead of ignore. Would that make using update as redundant?

So every time we update we would call the insert instead. Just wondering would that cause any confusion because update has the intention to update.

I guess if it removes some boiler plate code this should be better to use insert.

data/src/main/java/me/androidbox/data/remote/model/request/EventRequestModel.kt
steve1rm: @philipplackner 
The naming convention I think I read from some coding standard. However, I would like to stay with it for consistency.

data/src/main/java/me/androidbox/data/local/DatabaseConstant.kt
```diff
@@ -0,0 +1,11 @@
...
+
+object DatabaseConstant {
+    const val AGENDA_TABLE = "agenda_table"
+    const val FULL_AGENDA = "full_agenda"
+    const val ATTENDEE_TABLE = "attendee_table"
+    const val EVENT_TABLE = "event_table"
+    const val PHOTO_TABLE = "photo_table"
+    const val REMINDER_TABLE = "reminder_table"
+    const val TASK_TABLE = "task_table"
+}
```
philipplackner: This still contains the full agenda and agenda table which isn't needed

data/src/main/java/me/androidbox/data/local/dao/AttendeeDao.kt
```diff
@@ -0,0 +1,21 @@
...
+    fun getAttendee(): Flow<List<AttendeeEntity>>
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun insertAttendee(attendeeModel: AttendeeEntity)
+
+    @Delete
+    fun deleteAttendee(attendeeModel: AttendeeEntity)
+
+    @Query("DELETE FROM ${DatabaseConstant.ATTENDEE_TABLE}")
+    fun deleteAllAttendee()
```
philipplackner: I'd say you rather need a function to delete all attendees part of an event, but not all of them right? 🤔

data/src/main/java/me/androidbox/data/local/converter/AttendeeConverter.kt
```diff
@@ -0,0 +1,22 @@
...
+    @TypeConverter
+    fun fromJson(json: String): List<AttendeeEntity> {
+        return jsonAdapter.fromJson(json) ?: emptyList()
+    }
+
+    @TypeConverter
+    fun toJson(listOfAttendee: List<AttendeeEntity>): String {
+        return jsonAdapter.toJson(listOfAttendee)
+    }
+}
```
philipplackner: If you save attendees and photos as JSON, you don't need to make them separate tables of your database, they can simply be plain classes

data/src/main/java/me/androidbox/data/local/entity/PhotoEntity.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+import me.androidbox.data.local.DatabaseConstant
+
+@Entity(tableName = DatabaseConstant.PHOTO_TABLE)
+data class PhotoEntity(
+    /* TODO Can we use the key as a primary key as the key will be unique */
+    @PrimaryKey(autoGenerate = false)
+    val key: String,
```
philipplackner: Yes, you can 👍

philipplackner (approved):

data/src/main/java/me/androidbox/data/local/DatabaseConstant.kt
MKiperszmid: Here, Photo table and Attendee table are also optional, since you can save them as json

data/src/main/java/me/androidbox/data/local/converter/AttendeeConverter.kt
```diff
@@ -0,0 +1,22 @@
+package me.androidbox.data.local.converter
+
+import androidx.room.TypeConverter
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.Types
+import me.androidbox.data.local.entity.AttendeeEntity
+
+class AttendeeConverter {
+    private val moshi = Moshi.Builder().build()
```
MKiperszmid: You could receive the moshi in the constructor, and inject it with HIlt

data/src/main/java/me/androidbox/data/local/dao/AttendeeDao.kt
MKiperszmid: Just to clarify

You never want to delete ALL attendees from the DB, since the attendees are linked to different events
What you want, is to remove attendees from an event, but not from the DB

data/src/main/java/me/androidbox/data/local/dao/EventDao.kt
```diff
@@ -0,0 +1,21 @@
...
+    fun getEvent(): Flow<List<EventEntity>>
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun insertEvent(eventEntity: EventEntity)
+
+    @Delete
+    fun deleteEvent(eventEntity: EventEntity)
+
+    @Query("DELETE FROM ${DatabaseConstant.EVENT_TABLE}")
+    fun deleteAllEvent()
```
MKiperszmid, line 24: What would be the case where you'd want to remove all events ?

data/src/main/java/me/androidbox/data/local/dao/ReminderDao.kt
```diff
@@ -0,0 +1,21 @@
...
+    fun getReminder(): Flow<List<ReminderEntity>>
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun insertReminder(reminderEntity: ReminderEntity)
+
+    @Delete
+    fun deleteReminder(reminderEntity: ReminderEntity)
+
+    @Query("DELETE FROM ${DatabaseConstant.REMINDER_TABLE}")
+    fun deleteAllReminder()
```
MKiperszmid, line 24: Same here

Also, why did you add this method on Task and Event, but not on Reminder?

MKiperszmid (approved):
Couple comments

data/src/main/java/me/androidbox/data/local/converter/AttendeeConverter.kt
steve1rm: Yes, I was thinking of doing that.

data/src/main/java/me/androidbox/data/local/dao/AttendeeDao.kt
steve1rm: So this will be a query to delete all the attendee that have been added to that event.

data/src/main/java/me/androidbox/data/local/dao/EventDao.kt
```diff
@@ -0,0 +1,25 @@
...
+    fun getEvent(): Flow<List<EventEntity>>
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun insertEvent(eventEntity: EventEntity)
+
+    /** TODO
+     * Delete only a single event and all the attendees and photos that have been added to that event
+     * */
+    @Delete
+    fun deleteEvent(eventEntity: EventEntity)
```
steve1rm: Is this correct, we only want to delete a single event?

data/src/main/java/me/androidbox/data/local/dao/ReminderDao.kt
```diff
@@ -0,0 +1,25 @@
...
+    fun getReminder(): Flow<List<ReminderEntity>>
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun insertReminder(reminderEntity: ReminderEntity)
+
+    /**
+     * TODO Delete only a single reminder
+     * */
+    @Delete
+    fun deleteReminder(reminderEntity: ReminderEntity)
```
steve1rm: Is this correct that we only want to delete a single reminder

data/src/main/java/me/androidbox/data/local/dao/TaskDao.kt
```diff
@@ -0,0 +1,25 @@
...
+    fun getTask(): Flow<List<TaskEntity>>
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    fun insertTask(taskEntity: TaskEntity)
+
+    /** TODO
+     * Delete only a single task
+     * */
+    @Delete
+    fun deleteTask(taskEntity: TaskEntity)
```
steve1rm: Is this correct we only want to delete a single reminder

data/src/main/java/me/androidbox/data/local/dao/EventDao.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import androidx.room.*
+import kotlinx.coroutines.flow.Flow
+import me.androidbox.data.local.DatabaseConstant
+import me.androidbox.data.local.entity.EventEntity
+
+@Dao
+interface EventDao {
+    @Query("SELECT * FROM ${DatabaseConstant.EVENT_TABLE} where dateCreated = :selectedDate ORDER BY dateCreated ASC")
+    fun getEvent(selectedDate: Long): Flow<List<EventEntity>>
```
steve1rm: Would this be correct if we are going to query the event on the date it was created?

data/src/main/java/me/androidbox/data/local/dao/AttendeeDao.kt
philipplackner: I always use `OnConflictStrategy.REPLACE` since yes, it does remove the need of an `@Update` function. With that strategy, if you insert an event with an ID that already exists in the DB it will replace the existing event (and therefore update it). 
There's also a new `@Upsert` annotation for Room which does exactly that without needing to specify an OnConflictStrategy.

data/src/main/java/me/androidbox/data/local/dao/AttendeeDao.kt
philipplackner: If you save attendees as JSON part of an event, this whole DAO isn't necessary.

If you save attendees in a separate table and therefore need a DAO, yes it would be a query to delete all attendees part of an event with a given ID

data/src/main/java/me/androidbox/data/local/dao/EventDao.kt
philipplackner: Yes, but I'd usually make this a query that only takes an event's ID and deletes based on that. You don't always have the full instance of the event before deleting it.

data/src/main/java/me/androidbox/data/local/dao/ReminderDao.kt
philipplackner: Yes, same comment as for events

data/src/main/java/me/androidbox/data/local/dao/TaskDao.kt
philipplackner: Same as before

data/src/main/java/me/androidbox/data/local/dao/EventDao.kt
philipplackner: Explained that in WhatsApp

PR #10, opened 2023-02-28
[9] components module and loginScreen

steve1rm:
# Pull Request template

###  What does this PR solve?

- Adds components module 
- Adds general components, passwordEntry, UserInputEntry, TaskButton
- Adds LoginScreen
- Adds LoginViewModel

login


component/src/main/java/me/androidbox/component/general/PasswordEntry.kt
```diff
@@ -0,0 +1,112 @@
...
+import me.androidbox.component.R
+import me.androidbox.component.ui.theme.*
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun PasswordEntry(
+    modifier: Modifier = Modifier,
+    passwordValue: String,
+    placeholderText: String,
+    isPasswordVisible: MutableState<Boolean> = mutableStateOf(false),
```
philipplackner: States should be passed as a field directly to your composables and not as mutable state. This would very likely cause lots of different sources from where the state could be mutated which makes it super hard to find bugs. The only place where the state should be changed is in the corresponding ViewModel.

Therefore, here you should just pass a normal boolean whether the password is visible or not. The logic for toggling it belongs in the ViewModel.

component/src/main/java/me/androidbox/component/general/PasswordEntry.kt
```diff
@@ -0,0 +1,112 @@
...
+        trailingIcon = {
+            val visibilityIconId = if(isPasswordVisible.value) {
+                R.drawable.visible
+            } else {
+                R.drawable.hidden
+            }
+            Icon(
+                modifier = modifier.clickable {
+                    isPasswordVisible.value = !isPasswordVisible.value
+                },
```
philipplackner: Why do you pass the root modifier to this icon? If you make this text field stretch the whole width for example, this will cause the icon to also fill the whole width which is not what you want.
The passed modifier should therefore only apply to the very outer composable, in this case the text field.

component/src/main/java/me/androidbox/component/general/PasswordEntry.kt
```diff
@@ -0,0 +1,112 @@
...
+fun PasswordEntry(
+    modifier: Modifier = Modifier,
+    passwordValue: String,
+    placeholderText: String,
+    isPasswordVisible: MutableState<Boolean> = mutableStateOf(false),
+    onPasswordChange: (String) -> Unit,
+) {
+    OutlinedTextField(
+        modifier = modifier
+            .fillMaxWidth()
```
philipplackner: I'd not force the text field to fill the whole width in this composable. Imagine, you'd want to show 2 text fields next to each other. That wouldn't be possible with your implementation, since both would try to fill the whole width and the left one would push the other one out of the screen.
Where a composable is positioned and how large it is should be decided outside that composable at the screen level 👍

component/src/main/java/me/androidbox/component/general/PasswordEntry.kt
```diff
@@ -0,0 +1,112 @@
...
+        },
+        trailingIcon = {
+            val visibilityIconId = if(isPasswordVisible.value) {
+                R.drawable.visible
+            } else {
+                R.drawable.hidden
+            }
+            Icon(
+                modifier = modifier.clickable {
+                    isPasswordVisible.value = !isPasswordVisible.value
```
philipplackner: Here, instead of mutating the state, rather also expose a lambda when the user clicks on the password toggle

component/src/main/java/me/androidbox/component/general/TaskButton.kt
```diff
@@ -0,0 +1,64 @@
...
+
+@Composable
+fun TaskButton(
+    modifier: Modifier = Modifier,
+    buttonText: String,
+    textSize: TextUnit = 16.sp,
+    buttonTextColor: Color = MaterialTheme.colorScheme.LoginTextColor,
+    backgroundColor: Color = MaterialTheme.colorScheme.buttonColor,
+    onButtonClick: () -> Unit,
+    radius: Dp= 38.dp,
```
philipplackner: Maybe to be more explicit, I'd call this `borderRadius` since just `radius` sounds like this is a circular button

component/src/main/java/me/androidbox/component/general/TaskButton.kt
```diff
@@ -0,0 +1,64 @@
...
+    height: Dp = 50.dp
+) {
+    Button(
+        modifier = modifier
+            .height(height)
```
philipplackner: Same as before, you don't need to pass the height to this. Being able to pass a modifier from the outside already allows you to choose the height of this composable and flexibly choose any value for this.

component/src/main/java/me/androidbox/component/login/LoginScreen.kt
```diff
@@ -0,0 +1,138 @@
...
+        mutableStateOf("")
+    }
+
+    val isPasswordVisible = remember {
+        mutableStateOf(false)
+    }
+
+    var password by remember {
+        mutableStateOf("")
+    }
```
philipplackner: For all states in your app, I'd recommend using a ViewModel as stated in a previous comment. All these states won't survive configuration changes like screen rotations and will be reset after that. Using a ViewModel avoids that and gives you a central place to easily manage and test your state mapping logic

philipplackner (approved):

component/src/main/java/me/androidbox/component/general/PasswordEntry.kt
steve1rm: @philipplackner 
Thanks, I am new to compose so your insight here is very valuable.

presentation/src/main/java/me/androidbox/presentation/login/screen/LoginScreen.kt
```diff
@@ -0,0 +1,150 @@
...
+                    .padding(horizontal = 16.dp)
+            ) {
+                Spacer(modifier = modifier.height(50.dp))
+                UserInputEntry(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .background(
+                            shape = RoundedCornerShape(10.dp),
+                            color = MaterialTheme.colorScheme.backgroundInputEntry
+                        ),
```
steve1rm, line 66: @philipplackner @MKiperszmid 

So for modifiers, we should pass all modified properties i.e. `height`, `background`, `fullmaxWidth`, etc to the composable?

component/src/main/java/me/androidbox/component/general/PasswordEntry.kt
```diff
@@ -0,0 +1,107 @@
...
+        placeholder = {
+            Text(text = placeholderText, color = MaterialTheme.colorScheme.placeholderEntry)
+        },
+        trailingIcon = {
+            val visibilityIconId = if(isPasswordVisible) {
+                R.drawable.visible
+            } else {
+                R.drawable.hidden
+            }
+            Icon(
```
MKiperszmid: When you add Clickable to an Icon, it's recommended to use the composable "IconButton", since it already has the properties adapted to fit the thumb and some extra considerations

component/src/main/java/me/androidbox/component/general/PasswordEntry.kt
```diff
@@ -0,0 +1,107 @@
...
+                modifier = Modifier.clickable {
+                    onPasswordVisibilityClicked()
+                },
+                painter = painterResource(id = visibilityIconId),
+                contentDescription = "eye open close"
```
MKiperszmid: I would change this to "show/hide password"
Eye open close isn't really descriptive to people that require accessibility

component/src/main/java/me/androidbox/component/general/PasswordEntry.kt
```diff
@@ -0,0 +1,107 @@
...
+}
+
+@Composable
+@Preview(showBackground = true, name = "Shows password entry with viewable password")
+fun PreviewPasswordEntryVisible() {
+    BusbyTaskyTheme {
+        val (text, setText) = remember {
+            mutableStateOf("")
+        }
+
```
MKiperszmid: You don't need a mutablestate on this preview

You could just add default values

Ex:
passwordValue = "Password"
onPasswordChange = {}

component/src/main/java/me/androidbox/component/general/PasswordEntry.kt
```diff
@@ -0,0 +1,107 @@
...
+}
+
+@Composable
+@Preview(showBackground = true, name = "Shows password entry with viewable password")
+fun PreviewPasswordEntryHidden() {
+    BusbyTaskyTheme {
+
+        val (text, setText) = remember {
+            mutableStateOf("")
+        }
```
MKiperszmid: Same as my previous comment

component/src/main/java/me/androidbox/component/general/PasswordEntry.kt
```diff
@@ -0,0 +1,107 @@
...
+import androidx.compose.ui.text.input.PasswordVisualTransformation
+import androidx.compose.ui.text.input.VisualTransformation
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import me.androidbox.component.R
+import me.androidbox.component.ui.theme.*
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun PasswordEntry(
```
MKiperszmid: PasswordEntry isn't that descriptive in my opinion
I would go with something like PasswordTextField

component/src/main/java/me/androidbox/component/general/TaskButton.kt
```diff
@@ -0,0 +1,59 @@
...
+import androidx.compose.ui.unit.TextUnit
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import me.androidbox.component.ui.theme.BusbyTaskyTheme
+import me.androidbox.component.ui.theme.LoginTextColor
+import me.androidbox.component.ui.theme.buttonColor
+
+@Composable
+fun TaskButton(
+    modifier: Modifier = Modifier,
```
MKiperszmid: Modifier usually is the first default parameter

So in this case, buttonText should be first

component/src/main/java/me/androidbox/component/ui/theme/Color.kt
```diff
@@ -0,0 +1,60 @@
...
+val LightBlue = Color(0xFFB7C6DE)
+val Red = Color(0xFFFF7272)
+val Black = Color(0xFF16161C)
+val LightWhite = Color(0xFFEEF6FF)
+
+val ColorScheme.placeholderEntry: Color
+    @Composable
+    get() {
+        return LightGray
+    }
```
MKiperszmid, line 24: Couldn't this just be:
```
val ColorScheme.placeholderEntry = LightGray
```

presentation/src/main/java/me/androidbox/presentation/login/screen/LoginScreen.kt
```diff
@@ -0,0 +1,148 @@
...
+fun LoginScreen(
+    modifier: Modifier = Modifier,
+    loginViewModel: LoginViewModel = hiltViewModel()
+) {
+
+    Column(modifier = modifier
+        .fillMaxSize()
+        .background(Color.Black)) {
+
+        Spacer(modifier = modifier.height(48.dp))
```
MKiperszmid: You are reusing the main modifier on all composables. If you receive a modifier from parameter, you should use it on the Main Composable (In this case, the Column at line 36)

presentation/src/main/java/me/androidbox/presentation/login/screen/LoginScreen.kt
MKiperszmid: Exactly

I usually don't receive a Modifier on Screens. Since they are usually fillMaxSize() and such

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+class LoginViewModel @Inject constructor() : ViewModel() {
+
+    var username by mutableStateOf("")
+        private set
+
+    var password by mutableStateOf("")
+        private set
+
+    var isPasswordVisible by mutableStateOf(false)
+        private set
```
MKiperszmid, line 20: I personally like to keep my state in just one dataclass, since it makes it a bit more easier and cleaner

MKiperszmid (approved):
Couple of comments

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
steve1rm: @philipplackner @MKiperszmid 
For states I am wondering should we be using savedStatehandle with stateflow?

The reason I used mutableStateOf as its tried more closely to compose

component/src/main/java/me/androidbox/component/ui/theme/Color.kt
steve1rm: @MKiperszmid 

`val ColorScheme.placeholderEntry = LightGray`
After applying that change I got the following error:
`Extension property cannot be initialized because it has no backing field`

PR #11, opened 2023-02-28
[10] register screen

steve1rm:
# Pull Request template

###  What does this PR solve?
Creates registration screen
Creates RegistrationViewModel

registerscreen


component/src/main/java/me/androidbox/component/login/RegisterScreen.kt
```diff
@@ -0,0 +1,155 @@
...
+        mutableStateOf("")
+    }
+
+    val isPasswordVisible = remember {
+        mutableStateOf(false)
+    }
+
+    var password by remember {
+        mutableStateOf("")
+    }
```
philipplackner: Same as for the last PR, I'd use one ViewModel per screen

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/RegisterViewModel.kt
```diff
@@ -0,0 +1,40 @@
...
+        const val USERNAME = "username"
+        const val EMAIL_ADDRESS = "email_address"
+        const val PASSWORD = "password"
+        const val IS_PASSWORD_VISIBLE = "is_password_visible"
+    }
+
+    val username = savedStateHandle.getStateFlow(USERNAME, "")
+    val emailAddress = savedStateHandle.getStateFlow(EMAIL_ADDRESS, "")
+    val isPasswordVisible = savedStateHandle.getStateFlow(IS_PASSWORD_VISIBLE, false)
+    val password = savedStateHandle.getStateFlow(PASSWORD, "")
```
philipplackner, line 23: Yeah, that is exactly how you handle process death. As I said in WhatsApp, in most cases it's reasonable to not care about it. Here for example it's very likely the user wouldn't be very upset to re-enter their email after they came back to the app after 3h or so (they probably wouldn't even remember if they entered something or not). 
But there are definitely states, you really want to restore since they could mess up the UX on that screen.

philipplackner (approved):

PR #13, opened 2023-03-06
[11] Retrofit and authentication service endpoints

steve1rm:
# Pull Request template

###  What does this PR solve?

- Adds retrofit dependencies
- Adds authentication endpoints and service
- Adds interceptor for api key
- Adds base url to the buildConfig
- Adds api key to local properties


data/src/main/java/me/androidbox/data/di/NetworkModule.kt
```diff
@@ -0,0 +1,68 @@
...
+import me.androidbox.data.BuildConfig
+import me.androidbox.data.remote.interceptor.ApiKeyInterceptor
+import me.androidbox.data.remote.network.authentication.AuthenticationService
+import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
+import retrofit2.Retrofit
+import retrofit2.converter.moshi.MoshiConverterFactory
+
+@Module
+@InstallIn(SingletonComponent::class)
```
steve1rm, line 18: Is this correct that it should be singleton?

data/src/main/java/me/androidbox/data/remote/network/authentication/AuthenticationService.kt
```diff
@@ -0,0 +1,27 @@
...
+    @POST(EndPointConstant.LOGIN)
+    suspend fun login(
+        @Body loginRequestDto: LoginRequestDto
+    ) : LoginDto
+
+    @GET(EndPointConstant.LOGOUT)
+    suspend fun logout(): ErrorResponseDto
+
+    @GET(EndPointConstant.AUTHENTICATE)
+    suspend fun authenticate(): ErrorResponseDto
```
steve1rm: For this get response should I be returning the `ErrorResponseDto`

data/src/main/java/me/androidbox/data/remote/network/authentication/AuthenticationService.kt
```diff
@@ -0,0 +1,27 @@
...
+    suspend fun register(
+        @Body registrationRequestDto: RegistrationRequestDto)
+
+    @POST(EndPointConstant.LOGIN)
+    suspend fun login(
+        @Body loginRequestDto: LoginRequestDto
+    ) : LoginDto
+
+    @GET(EndPointConstant.LOGOUT)
+    suspend fun logout(): ErrorResponseDto
```
steve1rm: For this get response should I be returning the `ErrorResponseDto`

data/src/main/java/me/androidbox/data/di/NetworkModule.kt
```diff
@@ -0,0 +1,68 @@
...
+    @Provides
+    fun provideRetrofit(okHttpClient: OkHttpClient): Retrofit {
+        return Retrofit.Builder()
+            .baseUrl(BuildConfig.BUSBY_TASKY_API)
+            .client(okHttpClient)
+            .addConverterFactory(MoshiConverterFactory.create())
+            .build()
+    }
+
+    @Reusable
```
steve1rm: Would it be best if these are `Reusable` as dagger will only create them if necessary. Or should they be `Singleton`?

data/src/main/java/me/androidbox/data/di/NetworkModule.kt
MKiperszmid: Yes!

data/src/main/java/me/androidbox/data/di/NetworkModule.kt
MKiperszmid: I personally use Singleton when im on a SingletonComponent

data/src/main/java/me/androidbox/data/remote/network/EndPointConstant.kt
```diff
@@ -0,0 +1,8 @@
+package me.androidbox.data.remote.network
+
+object EndPointConstant {
+    const val REGISTER = "register"
+    const val LOGIN = "login"
+    const val LOGOUT = "logout"
+    const val AUTHENTICATE =  "authenticate"
+}
```
MKiperszmid, line 8: I personally don't like having them like this. I usually prefer having the string directly in the api call (in this case, AuthenticationService)

So just: @POST("/register") and so on

data/src/main/java/me/androidbox/data/remote/network/authentication/AuthenticationService.kt
MKiperszmid: No need. It could just be a ResponseBody
Since you won't display an error

data/src/main/java/me/androidbox/data/remote/network/authentication/AuthenticationService.kt
MKiperszmid: Same as before. It could be a ResponseBody, and then parse the error if the response is error

MKiperszmid (approved):
Mostly Answers to questions

data/src/main/java/me/androidbox/data/remote/network/EndPointConstant.kt
steve1rm: @MKiperszmid @philipplackner 
I would normally prefer to have all the paths in all one place so I can see all the them in one place

PR #14, opened 2023-03-07
[12] Adding feature for registering a user

steve1rm:
# Pull Request template

###  What does this PR solve?

- Adding feature to register a new user.
- Adding feature to login existing user
- Adds use cases

If I try add a second duplicated user, I get a crash as follows.  This only happens for duplicated emails. 
However, I am trying to catch the exception in the ResgisterViewModel, however, its still crashes. Any ideas.

```
--> POST https://tasky.pl-coding.com/register
22:23:20.218 26990-27089  I  Content-Type: application/json; charset=UTF-8
22:23:20.218 26990-27089  I  Content-Length: 70
22:23:20.219 26990-27089  I  {"fullName":"JoeBlogs","email":"joe@gmail.com","password":"Password1"}
22:23:20.219 26990-27089  I  --> END POST (70-byte body)
22:23:20.458 26990-27089  D  tagSocket(103) with statsTag=0xffffffff, statsUid=-1
22:23:21.859 26990-27089  I  <-- 409 https://tasky.pl-coding.com/register (1639ms)
22:23:21.859 26990-27089  I  date: Tue, 07 Mar 2023 15:23:20 GMT
22:23:21.859 26990-27089  I  content-type: application/json
22:23:21.859 26990-27089  I  content-length: 52
22:23:21.859 26990-27089  I  server: nginx/1.20.0
22:23:21.860 26990-27089  I  {"message":"A user with that email already exists."}
22:23:21.860 26990-27089  I  <-- END HTTP (52-byte body)
22:23:21.863 26990-27087  E  FATAL EXCEPTION: DefaultDispatcher-worker-2
                             Process: me.androidbox.presentation, PID: 26990
                             retrofit2.HttpException: HTTP 409 
                             	at retrofit2.KotlinExtensions$await$2$2.onResponse(KotlinExtensions.kt:53)
                             	at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:161)
                             	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:519)
                             	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
                             	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
                             	at java.lang.Thread.run(Thread.java:1012)
                             	Suppressed: kotlinx.coroutines.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@dd439c4, Dispatchers.IO]
```

However, if I use withContext the crash doesn't happen

```
       withContext(Dispatchers.IO) {
            /* MOCK DATA */
            val registrationRequestDto = RegistrationRequestDto(
                fullName = "JoeBlogs",
                email = "joe@gmail.com",
                password = "Password1"
            )

            authenticationService.register(
                registrationRequestDto
            )
        }
```

### DIFF
--- a/data/src/main/java/me/androidbox/data/di/AuthenticationModule.kt
+++ b/data/src/main/java/me/androidbox/data/di/AuthenticationModule.kt
@@ -12,5 +12,5 @@ import me.androidbox.domain.authentication.remote.AuthenticationRepository
 interface AuthenticationModule {
 
     @Binds
-    fun bindsAuthenticationImp(authenticationImp: AuthenticationRepositoryImp): AuthenticationRepository
+    fun bindsAuthenticationRepositoryImp(authenticationRepositoryImp: AuthenticationRepositoryImp): AuthenticationRepository
 }
--- a/data/src/main/java/me/androidbox/data/di/AuthenticationModule.kt
+++ b/data/src/main/java/me/androidbox/data/di/AuthenticationModule.kt
@@ -4,13 +4,13 @@ import dagger.Binds
 import dagger.Module
 import dagger.hilt.InstallIn
 import dagger.hilt.android.components.ViewModelComponent
-import me.androidbox.data.remote.authentication.AuthenticationImp
-import me.androidbox.domain.authentication.remote.Authentication
+import me.androidbox.data.remote.authentication.AuthenticationRepositoryImp
+import me.androidbox.domain.authentication.remote.AuthenticationRepository
 
 @InstallIn(ViewModelComponent::class)
 @Module
 interface AuthenticationModule {
 
     @Binds
-    fun bindsAuthenticationImp(authenticationImp: AuthenticationImp): Authentication
+    fun bindsAuthenticationImp(authenticationImp: AuthenticationRepositoryImp): AuthenticationRepository
 }
--- a/data/src/main/java/me/androidbox/data/di/AuthenticationModule.kt
+++ b/data/src/main/java/me/androidbox/data/di/AuthenticationModule.kt
@@ -1,0 +1,16 @@
+package me.androidbox.data.di
+
+import dagger.Binds
+import dagger.Module
+import dagger.hilt.InstallIn
+import dagger.hilt.android.components.ViewModelComponent
+import me.androidbox.data.remote.authentication.AuthenticationImp
+import me.androidbox.domain.authentication.remote.Authentication
+
+@InstallIn(ViewModelComponent::class)
+@Module
+interface AuthenticationModule {
+
+    @Binds
+    fun bindsAuthenticationImp(authenticationImp: AuthenticationImp): Authentication
+}
--- a/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
+++ b/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
@@ -1,7 +1,9 @@
 package me.androidbox.data.remote.authentication
 
+import me.androidbox.data.remote.model.request.LoginRequestDto
 import me.androidbox.data.remote.model.request.RegistrationRequestDto
 import me.androidbox.data.remote.network.authentication.AuthenticationService
+import me.androidbox.domain.authentication.model.LoginModel
 import me.androidbox.domain.authentication.remote.Authentication
 import javax.inject.Inject
 
@@ -12,13 +14,25 @@ class AuthenticationImp @Inject constructor(
     override suspend fun registerUser(fullName: String, email: String, password: String) {
         /* MOCK DATA */
         val registrationRequestDto = RegistrationRequestDto(
-            fullName = "JoeBlogs",
-            email = "joe1@gmail.com",
-            password = "Password1"
+            fullName = fullName,
+            email = email,
+            password = password
         )
 
-        authenticationService.register(
-            registrationRequestDto
+        authenticationService.register(registrationRequestDto)
+    }
+
+    override suspend fun loginUser(email: String, password: String): LoginModel {
+        val loginRequestDto = LoginRequestDto(
+            email = email,
+            password = password
+        )
+        val loginDto = authenticationService.login(loginRequestDto)
+
+        return LoginModel(
+            token = loginDto.token,
+            userId = loginDto.userId,
+            fullName = loginDto.fullName
         )
     }
 }
--- a/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
+++ b/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
@@ -1,0 +1,29 @@
+package me.androidbox.data.remote.authentication
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import me.androidbox.data.remote.model.request.RegistrationRequestDto
+import me.androidbox.data.remote.network.authentication.AuthenticationService
+import me.androidbox.domain.authentication.remote.Authentication
+import javax.inject.Inject
+
+class AuthenticationImp @Inject constructor(
+    private val authenticationService: AuthenticationService,
+) : Authentication {
+
+    override suspend fun registerUser(fullName: String, email: String, password: String) {
+        CoroutineScope(Dispatchers.IO).launch {
+            /* MOCK DATA */
+            val registrationRequestDto = RegistrationRequestDto(
+                fullName = "JoeBlogs",
+                email = "joe@gmail.com",
+                password = "Password1"
+            )
+
+            authenticationService.register(
+                registrationRequestDto
+            )
+        }
+    }
+}
--- a/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
+++ b/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
@@ -1,9 +1,5 @@
 package me.androidbox.data.remote.authentication
 
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
 import me.androidbox.data.remote.model.request.RegistrationRequestDto
 import me.androidbox.data.remote.network.authentication.AuthenticationService
 import me.androidbox.domain.authentication.remote.Authentication
@@ -14,18 +10,15 @@ class AuthenticationImp @Inject constructor(
 ) : Authentication {
 
     override suspend fun registerUser(fullName: String, email: String, password: String) {
-        /* TODO Pass in the dispatcher in the constructor */
-        withContext(Dispatchers.IO) {
-            /* MOCK DATA */
-            val registrationRequestDto = RegistrationRequestDto(
-                fullName = "JoeBlogs",
-                email = "joe@gmail.com",
-                password = "Password1"
-            )
+        /* MOCK DATA */
+        val registrationRequestDto = RegistrationRequestDto(
+            fullName = "JoeBlogs",
+            email = "joe1@gmail.com",
+            password = "Password1"
+        )
 
-            authenticationService.register(
-                registrationRequestDto
-            )
-        }
+        authenticationService.register(
+            registrationRequestDto
+        )
     }
 }
--- a/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
+++ b/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
@@ -3,6 +3,7 @@ package me.androidbox.data.remote.authentication
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import me.androidbox.data.remote.model.request.RegistrationRequestDto
 import me.androidbox.data.remote.network.authentication.AuthenticationService
 import me.androidbox.domain.authentication.remote.Authentication
@@ -13,7 +14,8 @@ class AuthenticationImp @Inject constructor(
 ) : Authentication {
 
     override suspend fun registerUser(fullName: String, email: String, password: String) {
-        CoroutineScope(Dispatchers.IO).launch {
+        /* TODO Pass in the dispatcher in the constructor */
+        withContext(Dispatchers.IO) {
             /* MOCK DATA */
             val registrationRequestDto = RegistrationRequestDto(
                 fullName = "JoeBlogs",
--- a/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
+++ b/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
@@ -12,7 +12,6 @@ class AuthenticationImp @Inject constructor(
 ) : Authentication {
 
     override suspend fun registerUser(fullName: String, email: String, password: String) {
-        /* MOCK DATA */
         val registrationRequestDto = RegistrationRequestDto(
             fullName = fullName,
             email = email,
--- a/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationRepositoryImp.kt
+++ b/data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationRepositoryImp.kt
@@ -3,6 +3,7 @@ package me.androidbox.data.remote.authentication
 import kotlinx.coroutines.CancellationException
 import me.androidbox.data.remote.model.request.LoginRequestDto
 import me.androidbox.data.remote.model.request.RegistrationRequestDto
+import me.androidbox.data.remote.model.response.LoginDto
 import me.androidbox.data.remote.network.authentication.AuthenticationService
 import me.androidbox.domain.authentication.ResponseState
 import me.androidbox.domain.authentication.model.Login
@@ -14,6 +15,22 @@ class AuthenticationRepositoryImp @Inject constructor(
     private val authenticationService: AuthenticationService,
 ) : AuthenticationRepository {
 
+    /* TODO Move this to a utils file to be used in other classes */
+    suspend fun <T> checkResult(block: suspend () -> T): Result<T> {
+        return try {
+            return Result.success(block())
+        }
+        catch (httpException: HttpException) {
+            Result.failure(httpException)
+        }
+        catch (exception: Exception) {
+            if (exception is CancellationException) {
+                throw exception
+            }
+            Result.failure(exception)
+        }
+    }
+
     override suspend fun registerUser(fullName: String, email: String, password: String): ResponseState<Unit> {
         val registrationRequestDto = RegistrationRequestDto(
             fullName = fullName,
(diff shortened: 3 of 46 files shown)


data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
```diff
@@ -0,0 +1,29 @@
...
+                email = "joe@gmail.com",
+                password = "Password1"
+            )
+
+            authenticationService.register(
+                registrationRequestDto
+            )
+        }
+    }
+}
```
steve1rm: Is this the best way to launching a coroutine? By using CoroutineScope

steve1rm: I have updated this to use `withContext` so this crash doesn't happen now.
What is the difference to using CoroutineScrope or withContext?

domain/src/main/java/me/androidbox/domain/authentication/usecase/RegisterUseCase.kt
```diff
@@ -0,0 +1,5 @@
+package me.androidbox.domain.authentication.usecase
+
+fun interface RegisterUseCase {
+    suspend fun execute(fullName: String, email: String, password: String)
+}
```
MKiperszmid: There's no real benefit in having a usecase as an interface
Since you want to actually test the functionality of said usecase

data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
MKiperszmid: CoroutineScope is a way to launch a coroutine
In a viewmodel usually you use viewmodelScope

WithContext, simply changes the context the current code is
So if you are in a Dispatchers.IO (Used for IO Operations like Network and DB), then you can't modify UI. Since that UI needs to be modified in Dispatchers.Main. But, if you run Network or DB on Dispatchers.Main, then the app will freeze since you shouldn't run heavy operations on the Main thread

domain/src/main/java/me/androidbox/domain/authentication/remote/Authentication.kt
```diff
@@ -0,0 +1,5 @@
+package me.androidbox.domain.authentication.remote
+
+interface Authentication {
+    suspend fun registerUser(fullName: String, email: String, password: String)
+}
```
MKiperszmid: What's the use of this interface?

MKiperszmid (approved):
Questions

domain/src/main/java/me/androidbox/domain/authentication/usecase/RegisterUseCase.kt
steve1rm: @MKiperszmid the implementation is called `RegisterUseCaseImp`

I prefer to use interfaces so I can pass them into constructors and use `@Binds` for dagger injection. I find it easier to mock interfaces in unit tests.

domain/src/main/java/me/androidbox/domain/authentication/remote/Authentication.kt
steve1rm: @MKiperszmid The implementation is in the data layer. Is called `AuthenticationImp`.

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
```diff
@@ -30,4 +46,24 @@ class LoginViewModel @Inject constructor() : ViewModel() {
...
+                loginMutableState.value = NetworkResponseState.Failure(httpException)
+            }
+            catch (exception: Exception) {
+                if(exception is CancellationException) {
+                    throw Exception(exception.message)
+                }
+                loginMutableState.value = NetworkResponseState.Failure(exception)
+            }
+        }
+    }
```
steve1rm, line 50: 1) launch the coroutine in viewModelScope
2) Sets the Loading state
3) Waits for the response back
4) sets the loginModel if success
5) set the exception if failure

I am wondering where should I collect the `loginState` should it be in the `LoginScreen` or should it be in the `HomeActivity`?

steve1rm: This would be the same for the `RegisterViewModel` and `RegisterScreen`

data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
```diff
@@ -0,0 +1,37 @@
+package me.androidbox.data.remote.authentication
+
+import me.androidbox.data.remote.model.request.LoginRequestDto
+import me.androidbox.data.remote.model.request.RegistrationRequestDto
+import me.androidbox.data.remote.network.authentication.AuthenticationService
+import me.androidbox.domain.authentication.model.LoginModel
+import me.androidbox.domain.authentication.remote.Authentication
+import javax.inject.Inject
+
+class AuthenticationImp @Inject constructor(
```
MKiperszmid: Is this the repository?

If so, it's common practice to have the word Repository in it
Ex: AuthenticationRepositoryImpl

domain/src/main/java/me/androidbox/domain/authentication/remote/Authentication.kt
```diff
@@ -0,0 +1,8 @@
+package me.androidbox.domain.authentication.remote
+
+import me.androidbox.domain.authentication.model.LoginModel
+
+interface Authentication {
```
MKiperszmid: If this is the repository, I'd name it AuthenticationRepository

domain/src/main/java/me/androidbox/domain/authentication/usecase/RegisterUseCase.kt
MKiperszmid: Gotcha

I was saying, since usually you tend to want the usecase to be ran on the test also

Since that's how you test it's actually working

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
MKiperszmid: Since you are passing the viewmodel to the LoginScreen, then you can collectAsState on the LoginScreen

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationImp.kt
steve1rm: @MKiperszmid @philipplackner 
This is requesting and fetching from the API endpoint. What would the best naming be for this?

domain/src/main/java/me/androidbox/domain/authentication/remote/Authentication.kt
steve1rm: @MKiperszmid @philipplackner 
Same as above, this is requesting and fetching from the API endpoint

domain/src/main/java/me/androidbox/domain/authentication/usecase/RegisterUseCase.kt
philipplackner: Yeah, have to agree with Martin here. You never want to replace the implementation of a use case for your tests. They contain your app's business logic and already are isolated as much as possible. You only want to create fakes for your boundaries (so basically things that leave the context of your local app's environment such as remote APIs, local DBs, dealing with Android OS services like AlarmManager, etc.). You use none of these in a use case.

Abstractions are expensive, every abstraction you add, adds quite a lot of complexity and cognitive load to your project. General rule of thumb: If you'll only ever need one implementation of an abstraction, leave it away, it's not needed.

presentation/src/main/java/me/androidbox/presentation/NetworkResponseState.kt
```diff
@@ -0,0 +1,11 @@
+package me.androidbox.presentation
+
+sealed interface NetworkResponseState<out T> {
+    object Idle : NetworkResponseState<Nothing>
+
+    object Loading : NetworkResponseState<Nothing>
```
philipplackner: I don't have a strong opinion on this, but these 2 are rather optional. You can also just set `isLoading` to true before you execute your network call and set it to false afterwards which is the approach I always use.

presentation/src/main/java/me/androidbox/presentation/di/UseCaseModule.kt
```diff
@@ -0,0 +1,24 @@
...
+@Module
+interface UseCaseModule {
+    @Reusable
+    @Binds
+    fun bindsRegisterUseCaseImp(registerUseCaseImp: RegisterUseCaseImp): RegisterUseCase
+
+    @Reusable
+    @Binds
+    fun bindsLoginUseCaseImp(loginUseCaseImp: LoginUseCaseImp): LoginUseCase
+}
```
philipplackner, line 24: Not needed, if you get rid of the interface

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
philipplackner: Also, exception handling for data related exceptions should happen in your data layer (in the repo), not in the ViewModel. Here, you're effectively binding implementation details to your presentation layer, since the `HttpException` comes from Retrofit. Therefore, presentation knows what you use to retrieve data, which it shouldn't in clean architecture. 
The goal is that if an implementation detail changes (e.g. your lib to make network calls), you ONLY want to change things in the data layer. Here, you'd need to change presentation as well.

Also mapping everything to `NetworkResponseState` should happen in the repository, so all the VM needs to do is to check the result from the repo and update its state based on that.

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/RegisterViewModel.kt
```diff
@@ -2,14 +2,26 @@ package me.androidbox.presentation.login.viewmodel
...
 import javax.inject.Inject
 
 @HiltViewModel
 class RegisterViewModel @Inject constructor(
-    private val savedStateHandle: SavedStateHandle
+    private val savedStateHandle: SavedStateHandle,
+    private val registerUseCase: RegisterUseCase
 ): ViewModel() {
 
+    private val registerMutableState = MutableStateFlow<NetworkResponseState<Unit>>(NetworkResponseState.Idle)
```
philipplackner: Instead of an idle state for enums/sealed classes, I'd just make it nullable if it doesn't exist. That's exactly what nulls are used for.

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/RegisterViewModel.kt
```diff
@@ -37,4 +49,27 @@ class RegisterViewModel @Inject constructor(
...
+            }
+            catch(httpException: HttpException) {
+                registerMutableState.value = NetworkResponseState.Failure(httpException)
+            }
+            catch (exception: Exception) {
+                if(exception is CancellationException) {
+                    throw Exception()
+                }
+            }
+        }
```
philipplackner: Same feedback here, that's for the data layer

philipplackner (approved):

data/src/main/java/me/androidbox/data/di/AuthenticationModule.kt
```diff
@@ -0,0 +1,16 @@
...
+import dagger.hilt.android.components.ViewModelComponent
+import me.androidbox.data.remote.authentication.AuthenticationRepositoryImp
+import me.androidbox.domain.authentication.remote.AuthenticationRepository
+
+@InstallIn(ViewModelComponent::class)
+@Module
+interface AuthenticationModule {
+
+    @Binds
+    fun bindsAuthenticationImp(authenticationImp: AuthenticationRepositoryImp): AuthenticationRepository
```
MKiperszmid: I would rename it to: bindsAuthenticationRepository
Since that's what it actually binds

data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationRepositoryImp.kt
```diff
@@ -0,0 +1,66 @@
...
+        }
+        catch(httpException: HttpException) {
+             NetworkResponseState.Failure(httpException)
+        }
+        catch (exception: Exception) {
+            if(exception is CancellationException) {
+                throw Exception()
+            }
+            NetworkResponseState.Failure(exception)
+        }
```
MKiperszmid: This can be simplified to:

```
catch (e: CancellationException) {
    throw e
} catch (exception: Exception) {
    NetworkResponseState.Failure(exception)
}
```

MKiperszmid: Also, when it's a cancellation exception, don't throw a new Exception by doing **Exception()** . You should re throw the current exception by doing **throw e** (e being the exception variable name)

data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationRepositoryImp.kt
```diff
@@ -0,0 +1,66 @@
...
+        catch(httpException: HttpException) {
+            NetworkResponseState.Failure(httpException)
+        }
+        catch (exception: Exception) {
+            if(exception is CancellationException) {
+                throw Exception()
+            }
+            NetworkResponseState.Failure(exception)
+        }
```
MKiperszmid: Since you are going to repeat this many times, you could have a function that does the following

Receive a lambda in the parameters, return a generic Result, and then it does the try catch, where in the try it calls the lambda and returns the success, and the catch returns the Result.failure(e) (and if it's cancellation exception, it throws it)

domain/src/main/java/me/androidbox/domain/authentication/NetworkResponseState.kt
```diff
@@ -0,0 +1,9 @@
+package me.androidbox.domain.authentication
+
+sealed interface NetworkResponseState<out T> {
```
MKiperszmid: I wouldn't name it NetworkResponseState, since this is in the domain, and you don't want the domain to know exactly from where the response is (it can be db also)

I'd name it to something like
```
Resource
```

domain/src/main/java/me/androidbox/domain/authentication/model/LoginModel.kt
```diff
@@ -0,0 +1,7 @@
+package me.androidbox.domain.authentication.model
+
+data class LoginModel(
```
MKiperszmid: No need to add the word "Model" to the models
For stuff like DTO's and Entities, it makes sense as to know which is which. But for domain models, which are used across the whole app, it's not needed nor recommended

You could simply name it Login or something a bit more descriptive like UserInformation

domain/src/main/java/me/androidbox/domain/authentication/usecase/LoginUseCase.kt
```diff
@@ -0,0 +1,8 @@
+package me.androidbox.domain.authentication.usecase
+
+import me.androidbox.domain.authentication.NetworkResponseState
+import me.androidbox.domain.authentication.model.LoginModel
+
+fun interface LoginUseCase {
+    suspend fun execute(email: String, password: String): NetworkResponseState<LoginModel>
+}
```
MKiperszmid: I'm still not fully understanding why the usecase is an interface
Use cases can be a simply class and that's it, since you won't be implementing this interface on testing or anywhere else

If an interface is going to be implemented only once, then there's no need to have an interface

For example, Repository, you'll implement it twice. One on the production code, and one for testing
For use cases, this isn't going to happen

domain/src/main/java/me/androidbox/domain/authentication/usecase/imp/RegisterUseCaseImp.kt
```diff
@@ -0,0 +1,18 @@
...
+
+class RegisterUseCaseImp @Inject constructor(
+    private val authenticationRepository: AuthenticationRepository
+) : RegisterUseCase {
+
+    override suspend fun execute(fullName: String, email: String, password: String): NetworkResponseState<Unit> {
+        return authenticationRepository.registerUser(
+            fullName = fullName,
+            email = email,
+            password = password)
```
MKiperszmid, line 16: The ')' could be in a separate line

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
```diff
@@ -4,11 +4,23 @@ import androidx.compose.runtime.getValue
...
-class LoginViewModel @Inject constructor() : ViewModel() {
+class LoginViewModel @Inject constructor(
+    private val loginUseCase: LoginUseCase
+) : ViewModel() {
+
+    private val loginMutableState: MutableStateFlow<NetworkResponseState<LoginModel>?> = MutableStateFlow(null)
+    val loginState = loginMutableState.asStateFlow()
 
     var username by mutableStateOf("")
         private set
```
MKiperszmid: Why do you use Stateflow on one,a nd mutableStateOf on the other?

Also, username isn't a thing in any part of the app

Login uses email/password, and Registration uses email/password/full name

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
```diff
@@ -4,11 +4,23 @@ import androidx.compose.runtime.getValue
...
+import kotlinx.coroutines.launch
+import me.androidbox.domain.authentication.NetworkResponseState
+import me.androidbox.domain.authentication.model.LoginModel
+import me.androidbox.domain.authentication.usecase.LoginUseCase
 import javax.inject.Inject
 
 @HiltViewModel
-class LoginViewModel @Inject constructor() : ViewModel() {
+class LoginViewModel @Inject constructor(
+    private val loginUseCase: LoginUseCase
```
MKiperszmid, line 19: You are missing the validation's usecases
Will that be on another PR?

MKiperszmid (approved):

domain/src/main/java/me/androidbox/domain/authentication/usecase/LoginUseCase.kt
steve1rm: @MKiperszmid @philipplackner 
This has been a familiar pattern that I have followed. Basically, I prefer to program to an interface. One reason is that I prefer to inject the interface into a constructor and use @Binds to bind it to an implementation.  Its easier to mock interfaces than concrete implementations. But I guess this can be open to discussion. But I value your reviews to have a second opinion on this.

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
steve1rm: @MKiperszmid 
I guess I was just experimenting with both. Yes, I should be consistent

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
steve1rm: @MKiperszmid That will be added in a separate PR. As I want to keep the PRs shorts. So validation will have a PR on is own as its just a separate unit of work.

domain/src/main/java/me/androidbox/domain/authentication/NetworkResponseState.kt
steve1rm: @MKiperszmid 
Renames to ResponseState

data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationRepositoryImp.kt
steve1rm: @MKiperszmid 
I was thinking the same thing. I don't like to repeat myself either

PR #15, opened 2023-03-08
[11] agenda item

steve1rm:
# Pull Request template

###  What does this PR solve?

- Create the Agenda items card (events, tasks, reminders)
- Create optionButton

taskcard
remindercard
eventcard


component/src/main/java/me/androidbox/component/general/OptionButton.kt
```diff
@@ -0,0 +1,50 @@
...
+            tint = dotColor
+        )
+        Spacer(modifier = Modifier.width(3.dp))
+        Icon(
+            painter = painterResource(id = R.drawable.white_dot),
+            contentDescription = "option button 3",
+            tint = dotColor
+        )
+    }
+}
```
steve1rm, line 42: Is this the best way to create the options button?

component/src/main/java/me/androidbox/component/agenda/AgendaCardType.kt
```diff
@@ -0,0 +1,170 @@
...
+@Composable
+fun AgendaCard(
+    modifier: Modifier = Modifier,
+    agendaCardType: AgendaCardType,
+    isAgendaCompleted: Boolean = false,
+    title: String,
+    subtitle: String,
+    dateTimeInfo: String,
+    onMenuOptionClicked: () -> Unit
+) {
```
MKiperszmid: This is kinda messy
Since the first parameters should be parameters that don't have a default value, and then you have parameters with default value

So something like:

```
title: String,
subtitle: String,
dateTimeInfo: String,
agendaCardType: AgendaCardType,
isAgendaCompleted: Boolean = false,
onMenuOptionClicked: () -> Unit
```

The only exception, is if you want to have a lambda like onMenuOptionClicked, which would be at the end, just like you did

component/src/main/java/me/androidbox/component/agenda/AgendaCardType.kt
```diff
@@ -0,0 +1,170 @@
...
+            Box(modifier = Modifier
+                .fillMaxWidth()) {
+                Row(modifier = Modifier.padding(start = 16.dp, top = 16.dp).align(Alignment.TopStart),
+                    verticalAlignment = Alignment.CenterVertically) {
+                    val agendaStatusIcon = if (isAgendaCompleted) {
+                        painterResource(id = R.drawable.completed)
+                    } else {
+                        painterResource(id = R.drawable.incomplete)
+                    }
+                    Icon(painter = agendaStatusIcon, contentDescription = "Agenda status")
```
MKiperszmid: "Agenda status" could be a string resource

component/src/main/java/me/androidbox/component/general/OptionButton.kt
MKiperszmid: Personally I like using a DropdownMenu
So when the button get's clicked, then it shows the DropdownMenu

MKiperszmid (approved):

steve1rm:
@MKiperszmid I will implement the dropDownMenu

PR #17, opened 2023-03-09
[14] Navigation loginscreen registerscreen

steve1rm:
# Pull Request template

###  What does this PR solve?
Adds navigation for the LoginScreen and RegistrationScreen

Can register and successfully and then go back to the LoginScreen


presentation/src/main/java/me/androidbox/presentation/navigation/Screen.kt
```diff
@@ -0,0 +1,13 @@
+package me.androidbox.presentation.navigation
+
+sealed interface Screen {
+    val route: String
+}
```
MKiperszmidSF: This could be a sealed class Screen(val route: String)

And then you could just

object LoginScreen : NavigationRoute("login_screen")
object RegisterScreen : NavigationRoute("register_screen")

MKiperszmidSF (approved):

presentation/src/main/java/me/androidbox/presentation/navigation/Screen.kt
steve1rm: @MKiperszmidSF Changed to to using a sealed class

presentation/src/main/java/me/androidbox/presentation/login/screen/LoginScreen.kt
```diff
@@ -24,15 +27,38 @@ import me.androidbox.component.general.TaskButton
...
 fun LoginScreen(
     modifier: Modifier = Modifier,
     loginViewModel: LoginViewModel = hiltViewModel(),
-    onSignUpClicked: () -> Unit
+    onSignUpClicked: () -> Unit,
+    onLoginSuccess: () -> Unit
 ) {
 
+    val loginState = loginViewModel.loginState.collectAsState()
+
```
philipplackner: I'd avoid passing the ViewModel directly to screen composables. This has 2 major reasons:
1. You're making it harder to have a preview, since you'll need to provide a fake ViewModel instance and fakes for everything the ViewModel needs in its constructor.
2. UI tests are also a bit harder since you'll need to provide these fakes.

What I suggest instead:
```
// In your ViewModel
@HiltViewModel
class SampleViewModel @Inject constructor(): ViewModel() {
    private val _state = MutableStateFlow(SampleState())
    val state = _state.asStateFlow()

    fun onEvent(event: SampleEvent) {
        when(event) {
            is SampleEvent.Event1 -> // Handle event1
            is SampleEvent.Event2 -> // Handle event2
            // ...
        }
    }

    // ...
}

// In your NavHost
composable(Routes.sample) {
    val viewModel = hiltViewModel<SampleViewModel>()
    val state by viewModel.state.collectAsState()
    SampleScreen(
        state = state,
        onEvent = viewModel::onEvent
    )
}

// Your screen composable
@Composable
fun SampleScreen(
    state: SampleState,
    onEvent: (SampleEvent) -> Unit
) {
    // ...
    Button(onClick = { onEvent(SampleEvent.ClickedButtonX) } {
        Text("Click me")
    }
}
```

presentation/src/main/java/me/androidbox/presentation/login/screen/LoginScreen.kt
```diff
@@ -24,15 +27,38 @@ import me.androidbox.component.general.TaskButton
...
+                loginViewModel.saveCurrentUserDetails(status.data)
+                onLoginSuccess()
+            }
+            is ResponseState.Failure -> {
+                /* Failure */
+                Log.d("LOGIN","${status.error}")
+            }
+            else -> {
+                /* no-op */
+            }
```
philipplackner: I personally like to write `else -> Unit` in such case which makes it clear this is a no-op

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
```diff
@@ -0,0 +1,97 @@
...
+                    registerViewModel.onEmailAddress(newEmail)
+                },
+                onPasswordChanged = { newPassword ->
+                    registerViewModel.onPasswordChanged(newPassword)
+                },
+                onPasswordVisibilityChanged = {
+                    registerViewModel.onPasswordVisibilityChanged()
+                },
+                onRegisterUser = {
+                    registerViewModel.registerUser()
```
steve1rm: Is this ok to call like this? As the RegisterViewModel has the state of the fullName, email, and password we can just call this to register a user.

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
```diff
@@ -0,0 +1,97 @@
...
+                    navHostController.popBackStack()
+                },
+                onRegistrationSuccess = {
+                    /* Registration Success */
+                    navHostController.popBackStack()
+                },
+                username = registerViewModel.username.collectAsState().value,
+                email = registerViewModel.email.collectAsState().value,
+                password = registerViewModel.password.collectAsState().value,
+                isPasswordVisible = registerViewModel.isPasswordVisible.collectAsState().value,
```
steve1rm: Is this ok to do like this `registerViewModel.isPasswordVisible.collectAsState().value`

MKiperszmid: I personally like to have just 1 state, and then act based on it

Ej: RegisterState

presentation/src/main/java/me/androidbox/presentation/login/screen/LoginScreen.kt
```diff
@@ -17,22 +20,46 @@ import androidx.compose.ui.text.style.TextAlign
...
+    loginState: State<ResponseState<Login>?>,
+    onLoginSuccess: (login: Login) -> Unit,
+    onSignUpClicked: () -> Unit,
+    email: String,
+    onEmailChanged: (email: String) -> Unit,
+    password: String,
+    onPasswordChanged: (newPassword: String) -> Unit,
+    onPasswordVisibilityChanged: () -> Unit,
+    isPasswordVisible: Boolean,
+    onLoginUser: (email: String, password: String) -> Unit
```
MKiperszmid: To avoid having a ton of lambdas, you can have an "onEvent" method in your viewmodel, and send LoginEvents

```
sealed class LoginEvent {
    data class PasswordChange(val password: String) : LoginEvent()
    data class EmailChange(val email: String) : LoginEvent()
    object Login : LoginEvent()
// etc...
}
```

and on your viewmodel:
```
 fun onEvent(event: LoginEvent) {
        when (event) {
            is LoginEvent.EmailChange -> {
                state = state.copy(
                    email = event.email
                )
            }
            is LoginEvent.PasswordChange -> {
                state = state.copy(
                    password = event.password
                )
            }
// ETC...
        }
}
```

MKiperszmid: And regarding passing the loginState: State<ResponseState<Login>?>, email, password, and so on

You could have just 1 state, that has every state related to the screen
Ex:

``` 
data class LoginState(
    val email: String = "",
    val emailError: String? = null,
    val password: String = "",
    val passwordError: String? = null,
    val isLoggedIn: Boolean = false,
    val isLoading: Boolean = false
// etc...
)
```

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
```diff
@@ -0,0 +1,97 @@
...
+                    loginViewModel.onPasswordChanged(newPassword)
+                },
+                onPasswordVisibilityChanged = {
+                    loginViewModel.onPasswordVisibilityChanged()
+                },
+                isPasswordVisible = loginViewModel.isPasswordVisible,
+                onLoginUser = { email, password ->
+                    loginViewModel.loginUser(email, password)
+                }
+            )
```
MKiperszmid: Having the onEvent would save you from writing all of this

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
MKiperszmid: Yes

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
```diff
@@ -0,0 +1,86 @@
...
+                    /* Signup clicked, navigate to register screen */
+                    navHostController.navigate(route = Screen.RegisterScreen.route)
+                },
+                onLoginSuccess = { login ->
+                    loginViewModel.saveCurrentUserDetails(login)
+                    /* TODO Navigate to the agenda screen (not implemented yet) */
+                },
+                loginScreenState = loginViewModel.loginScreenState.collectAsState().value
+            )
+        }
```
steve1rm, line 43: For `onSignUpClicked` and `onLoginSuccess` I think I should keep them here as this is where I want to do the navigation to other screens.

presentation/src/main/java/me/androidbox/presentation/login/screen/RegisterScreen.kt
steve1rm, line 1: @MKiperszmid @MKiperszmidSF @philipplackner 
Currently making changes for this. WIP. 
The LoginScreen is completed

presentation/src/main/java/me/androidbox/presentation/login/screen/LoginScreen.kt
```diff
@@ -17,22 +20,41 @@ import androidx.compose.ui.text.style.TextAlign
...
+import me.androidbox.domain.authentication.ResponseState
+import me.androidbox.domain.authentication.model.Login
 
 @Composable
 fun LoginScreen(
     modifier: Modifier = Modifier,
-    loginViewModel: LoginViewModel = hiltViewModel(),
-    onSignUpClicked: () -> Unit
+    loginScreenEvent: (LoginScreenEvent) -> Unit,
+    loginResponseState: State<ResponseState<Login>?>,
```
MKiperszmid: This could be added in the same LoginScreenState

presentation/src/main/java/me/androidbox/presentation/login/screen/LoginScreen.kt
```diff
@@ -64,11 +90,11 @@ fun LoginScreen(
...
                         ),
-                    inputValue = loginViewModel.email,
+                    inputValue = loginScreenState.email,
                     isInputValid = false,
                     placeholderText = stringResource(R.string.email_address),
-                    onInputChange = { newUsername ->
-                        loginViewModel.onUsernameChanged(newUsername)
+                    onInputChange = { newEmail ->
+                        loginScreenEvent(LoginScreenEvent.OnEmailChanged(newEmail))
                     }
```
MKiperszmid:
```suggestion
                    onInputChange = loginScreenEvent(LoginScreenEvent.OnEmailChanged(it))
```

MKiperszmid: Same with the rest

presentation/src/main/java/me/androidbox/presentation/login/screen/RegisterScreen.kt
MKiperszmid: Here you could add the loginResponseState in this state

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/login/screen/LoginScreen.kt
philipplackner: You can't use lambdas like that 🤔

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
```diff
@@ -12,44 +9,50 @@ import kotlinx.coroutines.launch
...
         viewModelScope.launch {
-            loginMutableState.value = loginUseCase.execute(email, password)
+            loginResponseMutableState.value = loginUseCase.execute(email, password)
         }
     }
 
     fun saveCurrentUserDetails(login: Login) {
-        /* Will call use-case to save user login details */
+        /* Will call use-case to save user login details to shared preferences WIP */
     }
```
philipplackner: This function doesn't need to be called from the UI. You can just call it after login was successful in the ViewModel or directly in the repo when you get the token response

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
philipplackner: Yeah, as we talked about in today's call having one state class synergizes well with Compose projects. Not mandatory, but I don't see a reason why you would not want to do that.
Also, IMO this reads better:
```
val state by viewModels.state.collectAsState()
YourScreen(
    state = state
)
```

compared to `state = viewModel.state.collectAsState().value`

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
steve1rm: @philipplackner @MKiperszmid 
Would this be acceptable, as I would have to check to make sure the response is success first.
```
 private fun loginUser(email: String, password: String) {
        viewModelScope.launch {
            val loginResponseState = loginUseCase.execute(email, password)

            if (loginResponseState is ResponseState.Success) {
                saveCurrentUserDetails(loginResponseState.data)
            }

            loginScreenMutableState.value  = loginScreenState.value.copy(
                loginResponseState = loginResponseState)
        }
    }

    private fun saveCurrentUserDetails(login: Login) {
        /* Will call use-case to save user login details to shared preferences WIP */
    }
```

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
```diff
@@ -12,40 +9,56 @@ import kotlinx.coroutines.launch
...
 import javax.inject.Inject
 
 @HiltViewModel
 class LoginViewModel @Inject constructor(
     private val loginUseCase: LoginUseCase
 ) : ViewModel() {
 
-    private val loginMutableState: MutableStateFlow<ResponseState<Login>?> = MutableStateFlow(null)
-    val loginState = loginMutableState.asStateFlow()
+    private val loginScreenMutableState = MutableStateFlow(AuthenticationScreenState<Login>("", "", false, "",null))
```
MKiperszmid, line 21: Something I like to do, is have the default values in the State's class
So on AuthenticationScreenState I would add the default values there

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
MKiperszmid: Discussed on Whatsapp

PR #18, opened 2023-03-12
[15] encrypted shared preferences

steve1rm:
# Pull Request template

###  What does this PR solve?

- Adds encrypted shared preferences to save the userId, token, and fullName,
- Adds use-cases
- Changes the name of the domain model `Login` to `LoginUser` for better naming.


data/src/main/java/me/androidbox/data/remote/preference/PreferenceRepositoryImp.kt
```diff
@@ -0,0 +1,82 @@
...
+            .putString(FULL_NAME_KEY, fullName)
+            .apply()
+    }
+
+    override fun retrieveCurrentUserOrNull(): LoginUser? {
+        val token = sharedPreferences.getString(TOKEN_KEY, "")
+        val userId = sharedPreferences.getString(USER_ID_KEY, "")
+        val fullName = sharedPreferences.getString(FULL_NAME_KEY, "")
+
+        return if(token != null && userId != null && fullName != null) {
```
MKiperszmid: What happens if the token, userid or fullname are empty?

I'd personally return null also

So you could fix it with isEmptyOrNull

data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationRepositoryImp.kt
```diff
@@ -47,7 +47,7 @@ class AuthenticationRepositoryImp @Inject constructor(
         /* TODO Is this the best way to extract the value from the Result<T> */
         val loginDto = result.getOrNull()
         return if(loginDto != null) {
-            val login = Login(
+            val login = LoginUser(
```
philipplackner: How about `LoggedInUser` or `AuthenticatedUser`? That's how I called it, but up to you ofc

domain/src/main/java/me/androidbox/domain/authentication/preference/PreferenceRepository.kt
```diff
@@ -0,0 +1,11 @@
+package me.androidbox.domain.authentication.preference
+
+import me.androidbox.domain.authentication.model.LoginUser
+
+interface PreferenceRepository {
+    fun saveToken(token: String)
+    fun saveUserId(userId: String)
+    fun saveFullName(fullName: String)
+    fun retrieveCurrentUserOrNull(): LoginUser?
```
MKiperszmid: I'm not a fan of adding a "OrNull" to a method
Usually, when the return is Nullable like in this case "LoginUser?" it implies that if no user is found, it will return null

data/src/main/java/me/androidbox/data/remote/preference/PreferenceRepositoryImp.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    private val sharedPreferences by lazy {
+        EncryptedSharedPreferences.create(
+            context,
+            "secret_shared_preferences",
+            masterKey,
+            EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
+            EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM
+        )
+    }
```
philipplackner: What if you'd need these prefs somewhere else? You'd need to create a new instance (which I'm not sure how heavy that is for encrypted prefs). Instead, I'd prefer injecting it via Hilt

domain/src/main/java/me/androidbox/domain/authentication/usecase/SaveFullNameUseCase.kt
```diff
@@ -0,0 +1,5 @@
+package me.androidbox.domain.authentication.usecase
+
+fun interface SaveFullNameUseCase {
```
MKiperszmid: Why are you saving it separately?

You could just have 1 method being "SaveUserInformation" or something similar, that just saves all 3 attributes at once

domain/src/main/java/me/androidbox/domain/authentication/usecase/imp/DeleteCurrentUserUseCaseImp.kt
```diff
@@ -0,0 +1,14 @@
+package me.androidbox.domain.authentication.usecase.imp
+
+import me.androidbox.domain.authentication.preference.PreferenceRepository
+import me.androidbox.domain.authentication.usecase.DeleteCurrentUserUseCase
+import javax.inject.Inject
+
+class DeleteCurrentUserUseCaseImp @Inject constructor(
+    private val preferenceRepository: PreferenceRepository
+) : DeleteCurrentUserUseCase {
```
philipplackner, line 9: There are zero reasons to abstract a use case. A use case is intended to contain business logic which you always want to test. There's no case in which you'd want to replace your real app's business logic with fake logic. 

General rule: If you only need a single implementation of something, don't abstract it. Just leads to tons of boilerplate code

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
```diff
@@ -18,7 +18,7 @@ class LoginViewModel @Inject constructor(
     private val loginUseCase: LoginUseCase
 ) : ViewModel() {
 
-    private val loginScreenMutableState = MutableStateFlow(AuthenticationScreenState<Login>("", "", false, "",null))
+    private val loginScreenMutableState = MutableStateFlow(AuthenticationScreenState<LoginUser>("", "", false, "",null))
```
MKiperszmid: I like to initialize the default values on the actual class, or in this case, you could simply make it so initial value is null

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
philipplackner: I personally prefer to set the default values of the state in the state itself and not here in the ViewModel. That way you can also easily reset the state by doing `_state.update { AuthenticationScreenState() }` without needing to reset each field manually

philipplackner (approved):

domain/src/main/java/me/androidbox/domain/authentication/preference/PreferenceRepository.kt
steve1rm: @MKiperszmid I read a chapter in this book `Effective Kotlin, Best Practices` by Marcin Moskala. That is where I started to follow this pattern. What does @philipplackner think about this naming convention?

domain/src/main/java/me/androidbox/domain/authentication/usecase/SaveFullNameUseCase.kt
steve1rm: @MKiperszmid Updated to have a single method to save the current logged in user

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
steve1rm: @MKiperszmid Sets the default values on the state class

data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationRepositoryImp.kt
steve1rm: @philipplackner I went with the name `AuthenticatedUser` :blush:

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/LoginViewModel.kt
steve1rm: @philipplackner Updated to have default values

philipplackner (approved):

domain/src/main/java/me/androidbox/domain/authentication/preference/PreferenceRepository.kt
philipplackner: I don't have a strong opinion here. The internal Kotlin functions also partly use this pattern, for example when accessing a list at an index:
`list.getOrNull(0)`

I personally go with what @MKiperszmid said, but both are obvious to me and not wrong

data/src/main/java/me/androidbox/data/remote/preference/PreferenceRepositoryImp.kt
steve1rm: @updated

PR #19, opened 2023-03-13
[16] splash screen

steve1rm:
# Pull Request template

###  What does this PR solve?

- Creates splash screen
- Create viewmodel that will authenticate login in user
- navigate to login if token expired, navigate to agenda if token not expired (Not implemented yet)

**TODO:**
_Interceptor for token in next PR_


presentation/src/main/java/me/androidbox/presentation/HomeActivity.kt
```diff
@@ -3,20 +3,47 @@ package me.androidbox.presentation
...
+            val authenticatedState = splashScreenViewModel.authenticationState.collectAsState()
+            val destination = when(authenticatedState.value) {
+                is ResponseState.Success -> {
+                    /* TODO Go to the Agenda Screen when implemented */
+                    Screen.RegisterScreen.route
+                }
+                else -> {
+                    Screen.LoginScreen.route
+                }
+            }
```
steve1rm, line 40: @philipplackner and @MKiperszmid Is this the best way to use this splashscreen? Everything works as expected. But I just want to make sure.

domain/src/main/java/me/androidbox/domain/authentication/usecase/imp/AuthenticateUserUseCaseImp.kt
```diff
@@ -0,0 +1,14 @@
...
+import me.androidbox.domain.authentication.usecase.AuthenticateUserUseCase
+import javax.inject.Inject
+
+class AuthenticateUserUseCaseImp @Inject constructor(
+    private val authenticationRepository: AuthenticationRepository)
+    : AuthenticateUserUseCase {
+    override suspend fun execute(): ResponseState<Unit> {
+        return authenticationRepository.authenticateUser()
+    }
+}
```
steve1rm, line 14: Still using the same usecase pattern. Will change this later.

data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationRepositoryImp.kt
```diff
@@ -57,4 +57,17 @@ class AuthenticationRepositoryImp @Inject constructor(
...
+    override suspend fun authenticateUser(): ResponseState<Unit> {
+        val result = checkResult<Unit> {
+            authenticationService.authenticate()
+        }
+
+        return if(result.isSuccess) {
+            ResponseState.Success(Unit)
+        }
+        else {
+            ResponseState.Failure(result.exceptionOrNull() as? Exception ?: Exception())
```
MKiperszmid, line 70: The as? Exception ?: Exception casting is weird

You could probably do:

```
val exception = result.exceptionOrNull() ?: Exception()
ResponseState.Failure(exception)

```

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/SplashScreenViewModel.kt
```diff
@@ -0,0 +1,30 @@
...
+import dagger.hilt.android.lifecycle.HiltViewModel
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+import me.androidbox.domain.authentication.ResponseState
+import me.androidbox.domain.authentication.usecase.AuthenticateUserUseCase
+import javax.inject.Inject
+
+@HiltViewModel
+class SplashScreenViewModel @Inject constructor(
```
MKiperszmid: I wouldn't call it SplashScreenViewModel, since authentication of a user isn't related to the splashscreen

I would call it MainViewModel or something similar. So it's more related to the MainActivity (in this case, HomeActivity)

presentation/src/main/java/me/androidbox/presentation/HomeActivity.kt
```diff
@@ -3,20 +3,47 @@ package me.androidbox.presentation
...
         super.onCreate(savedInstanceState)
+        val splashScreenViewModel by viewModels<SplashScreenViewModel>()
+
+        installSplashScreen().apply {
+            this.setKeepOnScreenCondition {
+                splashScreenViewModel.isCompletedState.value
+            }
+        }
+
+        splashScreenViewModel.authenticateUser()
```
MKiperszmid: Couldn't this be called directly on the init {} of the VM?

presentation/src/main/java/me/androidbox/presentation/HomeActivity.kt
MKiperszmid: Looks good, yes

presentation/src/main/java/me/androidbox/presentation/HomeActivity.kt
```diff
@@ -3,20 +3,47 @@ package me.androidbox.presentation
...
 @AndroidEntryPoint
 class HomeActivity : ComponentActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        val splashScreenViewModel by viewModels<SplashScreenViewModel>()
+
+        installSplashScreen().apply {
+            this.setKeepOnScreenCondition {
+                splashScreenViewModel.isCompletedState.value
```
MKiperszmid: This is a bit picky but

There's not really a need for isCompletedState

You can simply use something like:
authenticationState.value == null

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/remote/authentication/AuthenticationRepositoryImp.kt
steve1rm: @MKiperszmid The reason for the cast as exceptionOrNull return a Throwable and the Failure(...) only excepts a Exception. So I had to cast it to Throwable. 

However,  I have update the responseState to except a Throwable i.e. 

```
sealed interface ResponseState<out T> {
    object Loading : ResponseState<Nothing>

    data class Success<T>(val data: T) : ResponseState<T>

    data class Failure(val error: Throwable) : ResponseState<Nothing>
}
```

presentation/src/main/java/me/androidbox/presentation/login/viewmodel/SplashScreenViewModel.kt
steve1rm: @MKiperszmid Renamed to `HomeViewModel` as its related to the `HomeActivity`.

presentation/src/main/java/me/androidbox/presentation/HomeActivity.kt
steve1rm: @MKiperszmid Move that call in the init {...} of the HomeViewModel

presentation/src/main/java/me/androidbox/presentation/HomeActivity.kt
steve1rm: @MKiperszmid Removed the isCompleteState and replaced with your condition.

PR #20, opened 2023-03-15
[17] token interceptor

steve1rm:
# Pull Request template

###  What does this PR solve?

- Adds Token Interceptor


data/src/main/java/me/androidbox/data/remote/interceptor/TokenInterceptor.kt
```diff
@@ -0,0 +1,28 @@
+package me.androidbox.data.remote.interceptor
+
+import me.androidbox.domain.authentication.preference.PreferenceRepository
+import okhttp3.Interceptor
+import okhttp3.Response
+
+class TokenInterceptor(private val preferenceRepository: PreferenceRepository) : Interceptor {
+
+    private companion object {
+        const val AUTHORIZATION_HEADER = "Authorization"
```
philipplackner: This string will never change and this constant also doesn't add extra readability, so in that case I personally prefer to leave it

philipplackner (approved):

data/src/main/java/me/androidbox/data/remote/interceptor/TokenInterceptor.kt
```diff
@@ -0,0 +1,28 @@
...
+                .newBuilder()
+                .addHeader(AUTHORIZATION_HEADER, "Bearer ${loginUser.token}")
+                .build()
+
+            chain.proceed(request)
+        }
+        else {
+            chain.proceed(chain.request())
+        }
+    }
```
MKiperszmid: This is my personal opinion:

I usually prefer to not have != if later I'm going to add an else

So right now:

if (loginUser!= null) {} else {}

I prefer having a == null and then an else. 

So I would do:

```

return if (loginUser == null) {
    chain.proceed(chain.request())
} else {
    val request = chain.request()
                .newBuilder()
                .addHeader(AUTHORIZATION_HEADER, "Bearer ${loginUser.token}")
                .build()
    chain.proceed(request)
}
```

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/remote/interceptor/TokenInterceptor.kt
steve1rm: Updated to use the string directly

data/src/main/java/me/androidbox/data/remote/interceptor/TokenInterceptor.kt
steve1rm: Updated the to check for == null in the if statement

PR #21, opened 2023-03-17
[19] Agenda Screen and TopBar

steve1rm:
# Pull Request template

###  What does this PR solve?

- Adding the Agenda Screen
- Adds TopBar
- Opens Calendar when clicking the month
- Updates the month on the topbar
- Adds the user's initials to the profile button

agendascreen-with-calendar

### DIFF
--- a/component/build.gradle.kts
+++ b/component/build.gradle.kts
@@ -51,10 +51,6 @@ dependencies {
     implementation(libs.appcompat)
     implementation(libs.material)
 
-    implementation("com.maxkeppeler.sheets-compose-dialogs:core:1.1.1")
-    implementation("com.maxkeppeler.sheets-compose-dialogs:calendar:1.1.1")
-    coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:1.1.5")
-
     debugImplementation(tests.ui.tooling)
 
     testImplementation(tests.junit)
--- a/component/build.gradle.kts
+++ b/component/build.gradle.kts
@@ -26,6 +26,8 @@ android {
     }
 
     compileOptions {
+        isCoreLibraryDesugaringEnabled = true
+
         sourceCompatibility = JavaVersion.VERSION_1_8
         targetCompatibility = JavaVersion.VERSION_1_8
     }
@@ -48,6 +50,10 @@ dependencies {
     implementation(libs.core.ktx)
     implementation(libs.appcompat)
     implementation(libs.material)
+
+    implementation("com.maxkeppeler.sheets-compose-dialogs:core:1.1.1")
+    implementation("com.maxkeppeler.sheets-compose-dialogs:calendar:1.1.1")
+    coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:1.1.5")
 
     debugImplementation(tests.ui.tooling)
 
--- a/component/src/main/java/me/androidbox/component/agenda/AgendaTopBar.kt
+++ b/component/src/main/java/me/androidbox/component/agenda/AgendaTopBar.kt
@@ -1,6 +1,5 @@
 package me.androidbox.component.agenda
 
-import android.util.Log
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.*
@@ -27,6 +26,7 @@ import java.time.LocalDate
 @Composable
 fun AgendaTopBar(
     initials: String,
+    displayMonth: String,
     onProfileButtonClicked: () -> Unit,
     onCalendarClicked: () -> Unit,
     onDateChanged: (localDate: LocalDate) -> Unit,
@@ -55,7 +55,7 @@ fun AgendaTopBar(
             modifier = Modifier.clickable {
                 onCalendarClicked()
             },
-            text = "March",
+            text = displayMonth,
             fontSize = 16.sp,
             fontWeight = FontWeight.Bold,
             color = MaterialTheme.colorScheme.topbarFontColor
@@ -89,6 +89,7 @@ fun PreviewAgendaTopBar() {
                 .fillMaxWidth()
                 .background(color = MaterialTheme.colorScheme.backgroundColor)
                 .padding(horizontal = 16.dp),
+            displayMonth = "August",
             initials = "SM",
             onCalendarClicked = {},
             onProfileButtonClicked = {},
--- a/component/src/main/java/me/androidbox/component/agenda/AgendaTopBar.kt
+++ b/component/src/main/java/me/androidbox/component/agenda/AgendaTopBar.kt
@@ -3,7 +3,10 @@ package me.androidbox.component.agenda
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.*
-import androidx.compose.material3.*
+import androidx.compose.material3.Button
+import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -11,40 +14,19 @@ import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
-import com.maxkeppeker.sheets.core.models.base.rememberUseCaseState
-import com.maxkeppeler.sheets.calendar.CalendarDialog
-import com.maxkeppeler.sheets.calendar.models.CalendarConfig
-import com.maxkeppeler.sheets.calendar.models.CalendarSelection
-import com.maxkeppeler.sheets.calendar.models.CalendarStyle
 import me.androidbox.component.ui.theme.BusbyTaskyTheme
 import me.androidbox.component.ui.theme.backgroundColor
 import me.androidbox.component.ui.theme.topbarButtonBackgroundColor
 import me.androidbox.component.ui.theme.topbarFontColor
-import java.time.LocalDate
 
-@OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun AgendaTopBar(
     initials: String,
     displayMonth: String,
     onProfileButtonClicked: () -> Unit,
-    onCalendarClicked: () -> Unit,
-    onDateChanged: (localDate: LocalDate) -> Unit,
+    onDateClicked: () -> Unit,
     modifier: Modifier = Modifier,
 ) {
-    val calendarState = rememberUseCaseState()
-
-    CalendarDialog(
-        state = calendarState,
-        config = CalendarConfig(
-            style = CalendarStyle.MONTH,
-            monthSelection = true,
-            yearSelection = true
-        ),
-        selection = CalendarSelection.Date { localDate ->
-            onDateChanged(localDate)
-        }
-    )
 
     Row(
         modifier = modifier,
@@ -53,7 +35,7 @@ fun AgendaTopBar(
 
         Text(
             modifier = Modifier.clickable {
-                onCalendarClicked()
+                onDateClicked()
             },
             text = displayMonth,
             fontSize = 16.sp,
@@ -91,9 +73,8 @@ fun PreviewAgendaTopBar() {
                 .padding(horizontal = 16.dp),
             displayMonth = "August",
             initials = "SM",
-            onCalendarClicked = {},
+            onDateClicked = {},
             onProfileButtonClicked = {},
-            onDateChanged = {}
         )
     }
 }
--- a/component/src/main/java/me/androidbox/component/agenda/AgendaTopBar.kt
+++ b/component/src/main/java/me/androidbox/component/agenda/AgendaTopBar.kt
@@ -1,0 +1,98 @@
+package me.androidbox.component.agenda
+
+import android.util.Log
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.*
+import androidx.compose.material3.*
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.maxkeppeker.sheets.core.models.base.rememberUseCaseState
+import com.maxkeppeler.sheets.calendar.CalendarDialog
+import com.maxkeppeler.sheets.calendar.models.CalendarConfig
+import com.maxkeppeler.sheets.calendar.models.CalendarSelection
+import com.maxkeppeler.sheets.calendar.models.CalendarStyle
+import me.androidbox.component.ui.theme.BusbyTaskyTheme
+import me.androidbox.component.ui.theme.backgroundColor
+import me.androidbox.component.ui.theme.topbarButtonBackgroundColor
+import me.androidbox.component.ui.theme.topbarFontColor
+import java.time.LocalDate
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun AgendaTopBar(
+    initials: String,
+    onProfileButtonClicked: () -> Unit,
+    onCalendarClicked: () -> Unit,
+    onDateChanged: (localDate: LocalDate) -> Unit,
+    modifier: Modifier = Modifier,
+) {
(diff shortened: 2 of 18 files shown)


presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+import kotlinx.coroutines.flow.asStateFlow
+import me.androidbox.presentation.agenda.screen.AgendaScreenEvent
+import me.androidbox.presentation.agenda.screen.AgendaScreenState
+import javax.inject.Inject
+
+@HiltViewModel
+class AgendaViewModel @Inject constructor() : ViewModel() {
+
+    private val agendaScreenMutableState = MutableStateFlow(AgendaScreenState())
+    val agendaScreenState = agendaScreenMutableState.asStateFlow()
```
philipplackner: Any reason you don't stick to the general convention of just prepending an underscore to the mutable flow? (`_agendaScreenState`)
I haven't seen it be done differently anywhere

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    fun onAgendaScreenEvent(agendaScreenEvent: AgendaScreenEvent) {
+        when(agendaScreenEvent) {
+            is AgendaScreenEvent.OnDateChanged -> {
+                agendaScreenMutableState.value = agendaScreenState.value.copy(
+                    displayMonth = agendaScreenEvent.month
+                )
```
philipplackner: In general, I'd always use `_stateFlow.update { it.copy(...) }` to avoid race conditions. This is how it could happen:

*StateFlow holds value A*
- Thread 1 calls `copy()` on StateFlow with value A
- Before Thread 1 assigns its changed copy, Thread 2 updates StateFlow with value B
- Now, Thread 1 assigns its changed copy, which is still based on value A though, but the latest value is already value B. Therefore, value B would never be considered here.

Not always an issue, but I worked on highly stateful apps with frequently changing states, in which this really bit me, since such bugs are super hard to find.
Using `_state.value = _state.value.copy(...)` is fine if you always call this from a single thread (which you do here). But in ViewModels, you're very likely to not do that as soon as you'd change the state inside a `viewModelScope`.

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
```diff
@@ -65,5 +68,19 @@ fun NavigationGraph(
                 },
             )
         }
+
+        /* Agenda Screen */
+        composable(
+            route = Screen.AgendaScreen.route
+        ) {
+            val agendaViewModel: AgendaViewModel = hiltViewModel()
+            val agendaScreenState = agendaViewModel.agendaScreenState.collectAsState()
```
philipplackner: Recently the collectAsStateWithLifecycle() dependency got stable, so I'd recommend to use that since it's the safest way to collect flows in Android (currently, your flow will be collected even if the app is in the background). For that, you'll need this dependency:
```
implementation "androidx.lifecycle:lifecycle-runtime-compose:2.6.0"
```
Then use it like this:
```
val state by viewModel.state.collectAsStateWithLifecycle()
```

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
steve1rm: @philipplackner Updated to having a underscore for the postfix `_agendaScreenState`

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
steve1rm: @philipplackner Updated to use collectAsStateWithLifecycle for all collections in the navigationGraph

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
steve1rm: @philipplackner Updated to use the `.update { .... }`

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
```diff
@@ -0,0 +1,106 @@
...
+        state = agendaScreenState.calendarState,
+        config = CalendarConfig(
+            style = CalendarStyle.MONTH,
+            monthSelection = true,
+            yearSelection = true
+        ),
+        selection = CalendarSelection.Date { localDate ->
+            agendaScreenEvent(AgendaScreenEvent.OnDateChanged(localDate))
+        }
+    )
```
steve1rm: @philipplackner @MKiperszmid Is this the best place to have this calendar?

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenEvent.kt
```diff
@@ -0,0 +1,7 @@
+package me.androidbox.presentation.agenda.screen
+
+import java.time.LocalDate
+
+sealed interface AgendaScreenEvent {
+    data class OnDateChanged(val date: LocalDate) : AgendaScreenEvent
+}
```
steve1rm, line 7: @philipplackner @MKiperszmid Is it worth having a sealed interface for a single event? It could be possible we start with one then scale up to other events as the project grows. So I think it will be worth keeping it.

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenState.kt
```diff
@@ -0,0 +1,11 @@
+package me.androidbox.presentation.agenda.screen
+
+import com.maxkeppeker.sheets.core.models.base.UseCaseState
+import java.time.ZoneId
+import java.time.ZonedDateTime
+
+data class AgendaScreenState(
+    val displayMonth: String = ZonedDateTime.now(ZoneId.systemDefault()).month.toString(),
+    val usersInitials: String = "",
+    val calendarState: UseCaseState = UseCaseState()
```
steve1rm: @philipplackner @MKiperszmid This is the calendar state, is it ok to add it here?

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+        viewModelScope.launch {
+            preferenceRepository.retrieveCurrentUserOrNull()?.let { authenticatedUser ->
+                _agendaScreenState.update { agendaScreenState ->
+                    agendaScreenState.copy(
+                        usersInitials = authenticatedUser.fullName.extractInitials()
+                    )
+                }
+            }
+        }
+    }
```
steve1rm, line 51: @philipplackner @MKiperszmid After the user enters the Agenda screen fetching the user from shared preferences to show their initials in the top bar.

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -0,0 +1,61 @@
...
+        val listOfFullName = this.split(" ")
+        var initials = ""
+
+        listOfFullName.forEach { name ->
+            initials += name.first()
+        }
+
+        return initials.uppercase()
+    }
+}
```
steve1rm, line 52: @philipplackner @MKiperszmid This is a extension method that extracts the initials from the full name. An edge case could be in the user has a name that is more than 2 names. i.e. a middle name.
Should this be in a utility package or is it ok to have it here, if this is the only place we extract the initials.

component/src/main/java/me/androidbox/component/general/TaskActionButton.kt
```diff
@@ -0,0 +1,52 @@
...
+import androidx.compose.ui.unit.dp
+import me.androidbox.component.R
+import me.androidbox.component.ui.theme.BusbyTaskyTheme
+
+@Composable
+fun TaskActionButton(
+    @DrawableRes iconResource: Int,
+    onActionClicked: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
```
MKiperszmid, line 23: Whenever I have 1 lambda on a composable, I usually prefer it to be as the last parameter

So:

```
fun TaskActionButton(
    @DrawableRes iconResource: Int,
    onActionClicked: () -> Unit,
)
```

This is the only case where the optional parameter rule breaks (The rule where the optional parameters need to be after the obligatory parameters)

But only breaks with lambda. Why? Because kotlin has high-order functions

Basically, if you pass a lambda as last parameter, you can then do:

```
TaskActionButton(R.drawable.my_icon) {
// onActionClicked gets called here
} 

```

component/src/main/java/me/androidbox/component/general/TaskButton.kt
```diff
@@ -8,6 +8,7 @@ import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Shape
```
MKiperszmid: Are you using this import?

component/build.gradle.kts
```diff
@@ -49,6 +51,10 @@ dependencies {
     implementation(libs.appcompat)
     implementation(libs.material)
 
+    implementation("com.maxkeppeler.sheets-compose-dialogs:core:1.1.1")
+    implementation("com.maxkeppeler.sheets-compose-dialogs:calendar:1.1.1")
+    coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:1.1.5")
```
MKiperszmid: Are you using these on the component module?

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
MKiperszmid: Looks right, yes

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenEvent.kt
MKiperszmid: This screen will have more than just 1 event

Stuff like:
OnDayOfWeekChange
OnItemClick
OnNewItem

etc

So I would keep it, yes

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenState.kt
MKiperszmid: Personally since it's really UI dependant, I just leave it in the Composable

Since the only use for the state is to "show"

So what I do is:

Create the state on the composable. Ex:
val calendarState = rememberCalendarState() // Or whatever it's called

And then on the button click, I just:

calendarState.show()

No need for a state for it in this case

But that's my opinionated answer
@philipplackner might have a different answer

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
MKiperszmid: Yea that's correct

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
MKiperszmid: Opinionated answer:

I like to do the following
If I'm using it only once/twice, I have a Util class

If I use it in many places, then I have an extension function



Btw, this is a great method to unit test ;)

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenState.kt
steve1rm: @MKiperszmid Lets wait for @philipplackner opinion on this

component/src/main/java/me/androidbox/component/general/TaskButton.kt
steve1rm: @MKiperszmid Removed the unused import. Normally, I have a spotless check to would remove all unused imports but not had time to set it up in this project.

component/build.gradle.kts
steve1rm: @MKiperszmid Removed these. I first added the calendar but forgot to remove them.

component/src/main/java/me/androidbox/component/agenda/AgendaTopBar.kt
```diff
@@ -0,0 +1,80 @@
...
+            onClick = {
+                onProfileButtonClicked()
+            }) {
+            Text(
+                maxLines = 1,
+                text = initials,
+                fontSize = 13.sp,
+                fontWeight = FontWeight.SemiBold,
+            )
+        }
```
philipplackner, line 61: Shouldn't this initials bubble be round?

component/src/main/java/me/androidbox/component/general/TaskActionButton.kt
philipplackner: Personal opinion:
I only do this if the lambda describes content in form of a composable, such as for `Box`:
```
Box {
    // Other composables here
}
```
As soon as it's something that's not a composable function (like an onClick lambda), I *always* make it a named parameter. This not only improves readability but also avoids confusion because you'll mix non-Compose code with Compose code. For example:
```
Column {
    Row {
        TaskyActionButton(R.drawable.my_icon) {
             viewModel.doNetworkCall()
        }
    }
}
```
In this case, this looks like a side-effect, but it isn't. Also in a PR you won't be able to tell what this final lambda does (for a button you can guess it, but it's not always that clear).
This is much more readable:
```
Column {
    Row {
        TaskyActionButton(
            iconRes = R.drawable.my_icon,
            onClick = viewModel::doNetworkCall
        )
    }
}
```

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
philipplackner: Looks good, I personally put dialogs at the bottom of the screen composable, since when checking this screen, you likely care more often about the actual screen and not the dialogs. Putting it at the top, you first need to scroll down quite a bit if there are multiple dialogs

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenEvent.kt
philipplackner: Yeah, even if you'd only have one event, it's likely a screen gets more in future with changing requirements. It also adds some sort of consistency.

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenState.kt
philipplackner: Yeah any type of state you get via `rememberX` belongs in the UI, since getting that state instance is a composable function

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
philipplackner: Either way this is business logic and should go in domain therefore. Middle names are accepted and need to be handled here. In that case the initials would still be first name and last name's initials. 
Don't remember if you use use cases for your app, but if so, it should go in a use case. If not, it can be any type of util class

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenState.kt
steve1rm: I will keep the calendar state in the screen.

component/src/main/java/me/androidbox/component/general/TaskActionButton.kt
steve1rm: @philipplackner Will change to use named parameters

component/src/main/java/me/androidbox/component/agenda/AgendaTopBar.kt
steve1rm: @philipplackner This does create a round button with text inside for the initials. You can see the screenshot in the description.

PR #24, opened 2023-03-20
[22] inserting, deleting, fetching events

steve1rm:
# Pull Request template

###  What does this PR solve?

- Inserts events to local DB
- Deletes event based on the ID
- Fetches events between startDateTime and endDateTime


data/src/main/java/me/androidbox/data/local/dao/EventDao.kt
```diff
@@ -1,25 +1,21 @@
...
-import kotlinx.coroutines.flow.Flow
 import me.androidbox.data.local.DatabaseConstant
 import me.androidbox.data.local.entity.EventEntity
 
 @Dao
 interface EventDao {
-    @Query("SELECT * FROM ${DatabaseConstant.EVENT_TABLE} WHERE `from` >= :startTimeStamp AND `from` <= :endTimeStamp")
-    fun getEventsFromTimeStamp(startTimeStamp: Long, endTimeStamp: Long): Flow<List<EventEntity>>
+    @Query("SELECT * FROM ${DatabaseConstant.EVENT_TABLE} WHERE `startDateTime` >= :startTimeStamp AND `startDateTime` <= :endTimeStamp")
+    suspend fun getEventsFromTimeStamp(startTimeStamp: Long, endTimeStamp: Long): List<EventEntity>
```
MKiperszmid: Why did you stopped using Flows?

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/local/dao/EventDao.kt
steve1rm: @MKiperszmid I was having trouble with it. I don't think I was using it correctly. And I didn't want to spend too much time trying to fix it. However, I will come back to this later and maybe change it back to flows again.

philipplackner: Yeah I'd definitely make this observable in some form. If you now use `suspend` this will make the calling code quite different, so I'd recommend to go through this rather now. Not sure what went wrong for you, but you don't really need to do anything with this flow other than observing it in the UI

data/src/main/java/me/androidbox/data/local/event/EventRepositoryImp.kt
```diff
@@ -0,0 +1,95 @@
...
+                        startDateTime = eventEntity.startDateTime,
+                        endDateTime = eventEntity.endDateTime,
+                        remindAt = eventEntity.remindAt,
+                        host = eventEntity.host,
+                        isUserEventCreator = eventEntity.isUserEventCreator,
+                        attendees = eventEntity.attendees,
+                        photos = eventEntity.photos
+                    )
+                }
+
```
philipplackner: As in the previous PR, this should be a mapping function

data/src/main/java/me/androidbox/data/local/event/EventRepositoryImp.kt
```diff
@@ -0,0 +1,95 @@
...
+            title = event.title,
+            description = event.description,
+            startDateTime = event.startDateTime,
+            endDateTime = event.endDateTime,
+            remindAt = event.remindAt,
+            host = event.host,
+            isUserEventCreator = event.isUserEventCreator,
+            attendees = event.attendees,
+            photos = event.photos
+        )
```
philipplackner: Same here

domain/src/main/java/me/androidbox/domain/authentication/model/Event.kt
```diff
@@ -0,0 +1,14 @@
+package me.androidbox.domain.authentication.model
+
+data class Event(
+    val id: String,
+    val title: String,
+    val description: String,
+    val startDateTime: Long,
+    val endDateTime: Long,
+    val remindAt: Long,
+    val host: String, /* event creator ID */
```
philipplackner: If you feel like you need a comment to describe a field in a domain model, why not directly rename it to `eventCreatorId`?

philipplackner (approved):

domain/src/main/java/me/androidbox/domain/authentication/model/Event.kt
steve1rm: @philipplackner Updated to `eventCreatorId`

data/src/main/java/me/androidbox/data/mapper/DataToDomainMapper.kt
```diff
@@ -0,0 +1,5 @@
+package me.androidbox.data.mapper
+
+fun interface DataToDomainMapper<in E, out M> {
+    operator fun invoke(entity: E): M
+}
```
philipplackner, line 5: What's the purpose of abstracting a mapper? There's no real reason to do that. You can simply use extension functions, since you'll use them in your data layer only anyways:
```
fun EventEntity.toEvent(): Event { ... }

fun Event.toEventEntity(): EventEntity { ... }
```

data/src/main/java/me/androidbox/data/mapper/imp/DataToDomainMapperImp.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import me.androidbox.data.local.entity.EventEntity
+import me.androidbox.data.mapper.DataToDomainMapper
+import me.androidbox.domain.authentication.model.Event
+import javax.inject.Inject
+
+class DataToDomainMapperImp @Inject constructor()
+    : DataToDomainMapper<@JvmSuppressWildcards List<EventEntity>, @JvmSuppressWildcards List<Event>> {
+
+    override fun invoke(listOfEventEntity: List<EventEntity>): List<Event> {
```
philipplackner, line 11: Apart from my comment above that this is unnecessary, here you'd force the mapper to be called with a list of events. Sometimes, you just have a single event though, for example when going to the detail screen. Then this wouldn't work unless you pass a list with one item which doesn't seem very nice

philipplackner (approved):

data/src/main/java/me/androidbox/data/mapper/imp/DataToDomainMapperImp.kt
steve1rm: @philipplackner I am not using this anymore. I just forgot to remove it from the code base 😊

PR #25, opened 2023-03-21
[23] dropdown menu

steve1rm:
# Pull Request template

###  What does this PR solve?

- Create custom dropdownmenu that can display any number of options
- Create event and state for the string resources in the list that is passed to the dropdownmenu

dropdownmenu


presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
```diff
@@ -65,9 +66,16 @@ fun AgendaScreen(
...
+                    agendaScreenEvent(
+                        AgendaScreenEvent.OnShowDropdown(
+                            listOf(
+                                R.string.open,
+                                R.string.edit,
+                                R.string.delete
+                            )
+                        )
+                    )
+                })
```
steve1rm: @philipplackner @MKiperszmid Event to trigger opening the dropdownmenu which contains the string resources IDs to add to the dropdownmenu

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
```diff
@@ -92,6 +100,19 @@ fun AgendaScreen(
             agendaScreenEvent(AgendaScreenEvent.OnDateChanged(localDate))
         }
     )
+
+    AgendaDropDownMenu(
+        modifier = Modifier.background(color = MaterialTheme.colorScheme.dropDownMenuBackgroundColor),
+        shouldOpen = agendaScreenState.agendaDropdownItemId.isNotEmpty(),
```
steve1rm: @philipplackner @MKiperszmid When the state has a list of dropdown items (string resources) then show the dropdownmenu.

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
```diff
@@ -92,6 +100,19 @@ fun AgendaScreen(
...
+        onCloseDropdown = {
+            /* Empty the list of item ids so the isExpanded dropdown will close as the items are empty */
+            agendaScreenEvent(AgendaScreenEvent.OnShowDropdown(listOf()))
```
steve1rm: @philipplackner @MKiperszmid When the user clicked on the background screen. Set the state to an empty list so that the dropdownmenu will dismiss. 
i.e. `shouldOpen = agendaScreenState.agendaDropdownItemId.isNotEmpty()`

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
```diff
@@ -92,6 +100,19 @@ fun AgendaScreen(
...
+        },
+        listOfMenuItemId = listOf(R.string.open, R.string.edit, R.string.delete),
+        onSelectedOption = { item ->
+            Log.d("AGENDA", "ITEM [ $item ]")
+        }
+    )
 }
```
steve1rm: @philipplackner @MKiperszmid The dropdownmenu always displays on the left had of the screen. See picture in description. Is this because the AgendaDropdownMenu needs to be contained in a  Box or Column with the alignment set to `End`

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenState.kt
```diff
@@ -6,4 +6,5 @@ import java.time.ZonedDateTime
 data class AgendaScreenState(
     val displayMonth: String = ZonedDateTime.now(ZoneId.systemDefault()).month.toString(),
     val usersInitials: String = "",
+    val agendaDropdownItemId: List<Int> = listOf()
```
steve1rm: @MKiperszmid @philipplackner Adds the string resources that will be populate the drowdown menu

component/src/main/java/me/androidbox/component/general/AgendaDropDownMenu.kt
```diff
@@ -0,0 +1,95 @@
...
+import me.androidbox.component.R
+import me.androidbox.component.ui.theme.BusbyTaskyTheme
+import me.androidbox.component.ui.theme.divider
+import me.androidbox.component.ui.theme.dropDownMenuBackgroundColor
+import me.androidbox.component.ui.theme.dropDownMenuColor
+
+@Composable
+fun AgendaDropDownMenu(
+    shouldOpen: Boolean,
+    @StringRes listOfMenuItemId: List<Int>,
```
steve1rm: @philipplackner @MKiperszmid Do I need to annotate this with `@StringRes`?

MKiperszmid: I"m pretty sure this isn't possible

One possible solution could be to have a data class with just an id, and have that id annotated with @StringRes

MKiperszmid: Or one I personally like even more, is to have an ENUM with all the possible options

component/src/main/java/me/androidbox/component/general/AgendaDropDownMenu.kt
```diff
@@ -0,0 +1,95 @@
...
+                        fontSize = 16.sp,
+                        color = MaterialTheme.colorScheme.dropDownMenuColor)
+                },
+                onClick = {
+                    onSelectedOption(index)
+                }
+            )
+
+            /* Avoid displaying a divider for the final item */
+            if(index < listOfMenuItemId.count() - 1) {
```
MKiperszmid: Instead of
```
listOfMenuItemId.count() - 1
```

You could use 
```
listOfMenuItemId.lastIndex
```

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
MKiperszmid: Mmmm not fully understanding this

You are editing the list over an over just to show the dropdown

I personally recommend just having a Boolean in the state, and based on that, show or hide the dropdown

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
MKiperszmid: I'm not a big fan of this since you are adding and removing stuff from a list all the time

I would just have a Boolean that sets to true or false when it's clicked, and show the dropdown based on that
It's more clearer


I was only able to understand it, thanks to your comments in the PR
If I was just reading the code, it would make it really hard to understand

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
MKiperszmid: Why not simply using a Boolean instead of modifying the list?

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
MKiperszmid: One way is to do that. Have a box that fillMaxSize and set the content alignment to where you want

Another way, is to add a Dropdown to each item on the agenda. This would make it so the dropdown will always show at the right height on the screen

MKiperszmid (approved):

component/src/main/java/me/androidbox/component/general/AgendaDropDownMenu.kt
steve1rm: @MKiperszmid Updated

steve1rm: @MKiperszmid Removed the annotation `@StringRes`

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
steve1rm: @MKiperszmid updated

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
steve1rm: @MKiperszmid Updated

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
steve1rm: @MKiperszmid Updated to use boolean

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
steve1rm: @MKiperszmid I tried that by having a box to wrap the content in the scaffold. However, this didn't seem to work and the dropdown menu still displays on bottom start. Any ideas?

component/src/main/java/me/androidbox/component/general/AgendaDropDownMenu.kt
steve1rm: @MKiperszmid Removed the `@StringRes`

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
philipplackner: Something like this will show the dropdown above `FloatingActionButton`:
```
Box {
    DropdownMenu(...)
    FloatingActionButton(...)
}
```

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenState.kt
philipplackner: Why do these need to be part of the state? They never change

component/src/main/java/me/androidbox/component/general/AgendaDropDownMenu.kt
philipplackner: This would work: `listOfMenuItemId: List<@StringRes Int>`, but as my comment above this doesn't seem to be correct as a state

component/src/main/java/me/androidbox/component/general/AgendaDropDownMenu.kt
```diff
@@ -0,0 +1,95 @@
...
+            )
+
+            /* Avoid displaying a divider for the final item */
+            if(index < listOfMenuItemId.lastIndex) {
+                Divider(
+                    modifier = Modifier.fillMaxWidth(),
+                    color = MaterialTheme.colorScheme.divider,
+                    thickness = 1.dp
+                )
+            }
```
MKiperszmid: I'm pretty sure DropdownMenu already handles this (I might be wrong tho)

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
```diff
@@ -65,19 +65,51 @@ fun AgendaScreen(
...
-                    /* TODO Add dropdown menu to create new events, reminders, and tasks */
-                    Log.d("AGENDA_SCREEN", "Action button clicked")
-            })
+                    agendaScreenEvent(
+                        AgendaScreenEvent.OnShowDropdown(
+                            listOfItems = listOf(
+                                R.string.open,
+                                R.string.edit,
+                                R.string.delete
+                            ),
```
MKiperszmid: You are still changing the values on the dropdown here, and on line 104

component/src/main/java/me/androidbox/component/general/AgendaDropDownMenu.kt
```diff
@@ -0,0 +1,95 @@
...
+import me.androidbox.component.R
+import me.androidbox.component.ui.theme.BusbyTaskyTheme
+import me.androidbox.component.ui.theme.divider
+import me.androidbox.component.ui.theme.dropDownMenuBackgroundColor
+import me.androidbox.component.ui.theme.dropDownMenuColor
+
+@Composable
+fun AgendaDropDownMenu(
+    shouldOpenDropdown: Boolean,
+    listOfMenuItemId: List<Int>,
```
MKiperszmid, line 22: Why not use an Enum that handles all of this?

MKiperszmid (approved):

component/src/main/java/me/androidbox/component/general/AgendaDropDownMenu.kt
philipplackner: Yes it does, no need for this

philipplackner (approved):

component/src/main/java/me/androidbox/component/general/AgendaDropDownMenu.kt
steve1rm: @philipplackner Removed this from the state.

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreenState.kt
steve1rm: @philipplackner Removed them from the state

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
steve1rm: @philipplackner Thanks, its working now and it positions correctly.

PR #26, opened 2023-03-23
[24] event work manager

steve1rm:
# Pull Request template

###  What does this PR solve?

- Working Sending the multipart request for events using work manager


data/src/main/java/me/androidbox/data/local/converter/EventConverter.kt
```diff
@@ -0,0 +1,25 @@
...
+    @TypeConverter
+    fun fromJson(json: String): EventRequestDto? {
+        return jsonAdapter.fromJson(json)
+    }
+
+    @TypeConverter
+    fun toJson(eventRequestDto: EventRequestDto): String {
+        return jsonAdapter.toJson(eventRequestDto)
+    }
+}
```
philipplackner: TypeConverters are only used for Room. You don't need this to manually serialize an object for a network request

data/src/test/java/me/androidbox/data/local/converter/EventConverterTest.kt
```diff
@@ -0,0 +1,53 @@
...
+        val jsonString = "{\"id\":\"3061eb15-154b-4ea8-b89a-984cf16fa6d6\",\"title\":\"Title\",\"description\":\"description\",\"from\":3,\"to\":6,\"remindAt\":4,\"attendeeIds\":[\"25921c52-3e59-4acb-9c81-f79c8fbc2a27\",\"cd30b0a2-4015-43cb-b45d-70535bc31bb5\",\"8d1ed082-5859-4461-a1e6-268ff96d5ab9\",\"32ffd738-6a5f-4938-8f75-59302bd0443d\"]}"
+
+        // Act
+        val actual = eventConverter.fromJson(jsonString)
+
+        // Assert
+        Truth.assertThat(actual).isNotNull()
+        Truth.assertThat(actual?.id).isEqualTo("3061eb15-154b-4ea8-b89a-984cf16fa6d6")
+    }
+}
```
philipplackner: What you're testing here is nothing else than the Moshi library. You can be sure that the authors of that library already considered testing. You never really test other SDKs, but rather just your own code.

presentation/src/main/java/me/androidbox/presentation/HomeActivity.kt
```diff
@@ -21,13 +26,39 @@ class HomeActivity : ComponentActivity() {
...
+            
+            val downloadInfo = remember(key1 = workInfos) {
+                workInfos?.find { workInfo ->
+                    workInfo.id == uploadEventWorkerRequest.id
+                }
+            }
+
+            val eventState by derivedStateOf {
+                val downloadEvent = downloadInfo?.outputData?.getString("")?.toUri()
+            }
```
philipplackner: This `derivedStateOf` does nothing, since you don't return any state in the block (and if you do, it would need to be wrapped with `mutableStateOf()`)

presentation/src/main/java/me/androidbox/presentation/util/imp/CreatePhotoMultipartImp.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    private fun getFileExtensionFromUri(uri: Uri): String {
+        val mimeTypeMap = MimeTypeMap.getSingleton()
+        val extension = mimeTypeMap.getExtensionFromMimeType(context.contentResolver.getType(uri)) ?: ""
+
+        return extension.lowercase(Locale.getDefault())
+    }
+
+    private suspend fun getByteArrayFromUri(uri: Uri): ByteArray {
+       return withContext(Dispatchers.Default) {
```
philipplackner: Shouldn't this use the `IO` dispatcher since reading bytes is an IO operation?

presentation/src/main/java/me/androidbox/presentation/util/imp/CreatePhotoMultipartImp.kt
```diff
@@ -0,0 +1,65 @@
...
+content://com.android.providers.media.documents/document/image%3A38324
+content://com.google.android.apps.photos.contentprovider/0/1/mediakey%3A%2Flocal%253A44f502c3-7f36-4f9b-887c-5d8f2f45f3cc/ORIGINAL/NONE/image%2Fjpeg/430889281
+content://com.google.android.apps.photos.contentprovider/-1/1/content%3A%2F%2Fmedia%2Fexternal%2Fimages%2Fmedia%2F37452/ORIGINAL/NONE/image%2Fjpeg/640889962
+content://com.google.android.apps.photos.contentprovider/0/1/mediakey%3A%2Flocal%253A1e78ee30-5b74-4a1e-862b-a13525f26cd4/ORIGINAL/NONE/image%2Fjpeg/63330985
+content://com.google.android.apps.photos.contentprovider/0/1/content%3A%2F%2Fmedia%2Fexternal%2Fimages%2Fmedia%2F38330/ORIGINAL/NONE/image%2Fjpeg/1631907737
+content://com.android.providers.media.documents/document/image%3A38324
+content://com.android.providers.media.documents/document/image%3A38337
+* */
+
+class CreatePhotoMultipartImp @Inject constructor(@ApplicationContext private val context: Context) : CreatePhotoMultipart {
```
philipplackner: What do you gain by having an abstraction for this?

presentation/src/main/java/me/androidbox/presentation/worker/UploadEventWorker.kt
```diff
@@ -0,0 +1,105 @@
+package me.androidbox.presentation.worker
```
philipplackner: WorkManager workers are rather data related, especially if they upload something like here

presentation/src/main/java/me/androidbox/presentation/worker/UploadEventWorker.kt
```diff
@@ -0,0 +1,105 @@
...
+
+        /* Serialize the event request to json */
+        val eventRequestJson = eventConverter.toJson(eventRequest)
+
+        /* These are just a random set of photos that I got from selecting photos from the device.
+         * I just copied them here to try and send them as a multi-part request just to test if I can send them */
+        val listOfPhotos = listOf(
+            "content://com.google.android.apps.photos.contentprovider/0/1/mediakey%3A%2Flocal%253A44f502c3-7f36-4f9b-887c-5d8f2f45f3cc/ORIGINAL/NONE/image%2Fjpeg/430889281",
+            "content://com.android.providers.media.documents/document/image%3A38324",
+            "content://com.google.android.apps.photos.contentprovider/-1/1/content%3A%2F%2Fmedia%2Fexternal%2Fimages%2Fmedia%2F37452/ORIGINAL/NONE/image%2Fjpeg/640889962")
```
philipplackner: AFAIK, these may differ each time you retrieve them. Not tested, but have heard people say that. I'd test this with a Uri you really picked on the device (this could explain your `FileNotFoundException`).

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/HomeActivity.kt
steve1rm: @philipplackner I was just trying this out and forgot to remove it.

presentation/src/main/java/me/androidbox/presentation/worker/UploadEventWorker.kt
steve1rm: @philipplackner so this should belong in my Data Layer module. I did have it there ordinary but moved it to the presentation module because that is where the work would be triggered from.

presentation/src/main/java/me/androidbox/presentation/util/imp/CreatePhotoMultipartImp.kt
steve1rm: @philipplackner Changed to the IO dispatcher

data/src/main/java/me/androidbox/data/local/converter/EventConverter.kt
steve1rm: @philipplackner So I should be doing something like the following:
```
        /* Serialize the event request to json */
        val moshi = Moshi.Builder().build()
        val jsonAdapter = moshi.adapter<EventRequestDto>(EventRequestDto::class.java)
        val eventRequestJson = jsonAdapter.toJson(eventRequest)
```

data/src/test/java/me/androidbox/data/local/converter/EventConverterTest.kt
steve1rm: @philipplackner Removed the unit tests

presentation/src/main/java/me/androidbox/presentation/util/imp/CreatePhotoMultipartImp.kt
steve1rm: @philipplackner This is the part I am stuck on. Does this code look correct?

data/src/main/java/me/androidbox/data/local/converter/EventConverter.kt
philipplackner: Yes 👍

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/worker/UploadEventWorker.kt
philipplackner: No, the work could as well be triggered from your data layer in the repository. That's where the request is executed, the worker is just a tool to execute it reliably

MKiperszmid: A way to trigger it from Presentation, would be to abstract it in the Domain layer, and have the implementation on the data layer

Kinda like what you did for the Email Validator

presentation/src/main/java/me/androidbox/presentation/worker/UploadEventWorker.kt
```diff
@@ -0,0 +1,98 @@
...
+                listOfPhoto = listOfPhotoMultiPart,
+                eventBody = MultipartBody.Part.createFormData("create_event_request", eventRequestJson)
+            )
+        }
+
+        /* Check if the request was successful or not */
+        val result = responseResult.fold(
+            onSuccess = { eventDto ->
+                val outputData = workDataOf("eventKey" to eventDto)
+                Result.success(outputData)
```
steve1rm: @MKiperszmid @philipplackner Is this correct on how to pass data back from the work manager: I am getting this crash:
```
Work [ id=eb508899-6586-4674-94f7-8a17420983f0, tags={ me.androidbox.presentation.worker.UploadEventWorker } ] failed because it threw an exception/error
    java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: Key eventKeyhas invalid type class me.androidbox.data.remote.model.response.EventDto
```

steve1rm: Another question, what do we do with this EventDto object that is returned from the endpoint? I save the event into Room DB before I sending the request via work manager.

philipplackner: Yes it's correct, but you can only pass primitive types (and maybe Parcelables I guess). It just doesn't know how to pass your `EventDto` back as a result. You can try to make it a `Parcelable` and if that doesn't work pass a serialized JSON string of the DTO object.
Either way, I wonder why you need to pass the whole event back? Why not just the info that it was successful?

data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
```diff
@@ -0,0 +1,58 @@
...
+                IS_EDIT_MODE to isEditMode,
+                EVENT to eventRequestJson,
+                EVENT_PHOTOS to event.photos.toList()
+            )
+        }
+        else {
+            workDataOf(
+                IS_EDIT_MODE to isEditMode,
+                EVENT to eventRequestJson)
+        }
```
steve1rm: @philipplackner @MKiperszmid Not sure if this is worth doing. I am just thinking its not worth adding the photos if the user didn't select any photos.

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
```diff
@@ -0,0 +1,100 @@
...
+                )
+            }
+
+            val result = responseResult.fold(
+                onSuccess = { eventDto ->
+                    if(eventDto.photos.isNotEmpty()) {
+                        val outputData = photoToJson(eventDto)
+
+                        Result.success(outputData)
+                    }
```
steve1rm, line 65: @philipplackner @MKiperszmid Will return the photos url and keys in the Success, if there any photos that was uploaded when creating or updating an event. Is this correct?

data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
```diff
@@ -0,0 +1,58 @@
...
+
+        val uploadWorkerRequest  = OneTimeWorkRequestBuilder<UploadEventWorker>()
+            .setConstraints(Constraints(Constraints.Builder().setRequiredNetworkType(NetworkType.CONNECTED).build()))
+            .setInputData(inputData = eventInputData).build()
+
+        workManager.enqueue(uploadWorkerRequest)
+    }
```
steve1rm, line 54: @MKiperszmid @MKiperszmid What is the best way to handle if there is no internet?
i.e. The user has lost connection, this will fail. Would the user have to click save again to upload the event if the first time it fails.

What happens if the there is no internet connection so the upload will fail. But if the connection is suddenly restored what will be the best way to handle that case?

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
```diff
@@ -0,0 +1,100 @@
...
+        val eventRequestJson = inputData.getString(UploadEventImp.EVENT)
+        val isEditMode = inputData.getBoolean(UploadEventImp.IS_EDIT_MODE, false)
+
+        val formData = getFormData(isEditMode)
+
+        if(!eventRequestJson.isNullOrEmpty()) {
+            val listOfPhotoMultiPart = listOfPhoto?.let {
+                /* Create the multipart for the list of photos */
+                createPhotoMultipart.createMultipartPhotos(listOfPhoto)
+            } ?: listOf()
```
steve1rm, line 46: @philipplackner @MKiperszmid when setting the parameters in the `UpLoadEventImp` I added a check to only add the photos to the inputData if the photo list wasn't empty. So I guess having this check here would mean its not worth checking as the inputData would be null if there wasn't any added photos in the first place.

data/src/main/java/me/androidbox/data/remote/network/EndPointConstant.kt
```diff
@@ -5,4 +5,5 @@ object EndPointConstant {
     const val LOGIN = "login"
     const val LOGOUT = "logout"
     const val AUTHENTICATE =  "authenticate"
+    const val EVENT = "event"
```
MKiperszmid, line 8: You could just have a constant on each of the service
There's no added benefit on having an extra class

It's not like the constants will be shared between 2 or more classes

data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
MKiperszmid: Tbh this is extra code that doesn't really do much
If you send an empty photo list, backend will not do anything 
So it can be removed

data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
```diff
@@ -0,0 +1,58 @@
...
+                EVENT_PHOTOS to event.photos.toList()
+            )
+        }
+        else {
+            workDataOf(
+                IS_EDIT_MODE to isEditMode,
+                EVENT to eventRequestJson)
+        }
+
+        val uploadWorkerRequest  = OneTimeWorkRequestBuilder<UploadEventWorker>()
```
MKiperszmid, line 45: What if the upload fails? You could add some backoff time 

Check "setBackoffCriteria"

data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
MKiperszmid: If i'm not mistaken, WorkManager should automatically handle it

Adding a backoff (previous comment) will make it so once there's connection, it will retry again

Tho we can wait for @philipplackner to fully confirm this

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
MKiperszmid: I would remove the other stuff and just have this here

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
MKiperszmid: Yea but the question is, do you really care about the response of the Success?

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
```diff
@@ -0,0 +1,100 @@
...
+                        val outputData = photoToJson(eventDto)
+
+                        Result.success(outputData)
+                    }
+                    else {
+                        Result.success()
+                    }
+                },
+                onFailure = { throwable ->
+                    val errorDate = workDataOf(ERROR to throwable)
```
MKiperszmid: Why is it named
errorDate?

data/src/main/java/me/androidbox/data/worker_manager/util/CreatePhotoMultipart.kt
```diff
@@ -0,0 +1,7 @@
+package me.androidbox.data.worker_manager.util
+
+import okhttp3.MultipartBody
+
+interface CreatePhotoMultipart {
```
MKiperszmid, line 5: Do you really need this interface?

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -146,17 +150,28 @@ class EventViewModel @Inject constructor(
...
                 Attendee(2, "gmail", "peter rab", UUID.randomUUID().toString(), UUID.randomUUID().toString(), false, 2L)),
             photos = eventScreenState.value.listOfPhotoUri
         )
 
         viewModelScope.launch {
-            eventRepository.insertEvent(event)
-
-            val alarmItem = event.toAlarmItem(AgendaType.EVENT)
-            alarmScheduler.scheduleAlarmReminder(alarmItem)
+            when(val responseState = eventRepository.insertEvent(event)) {
```
MKiperszmid, line 161: In what cases will this ever return something different than success?
Since you are just inserting to a db

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
steve1rm: @MKiperszmid I can remove it if there is nothing that can go wrong with an insert. @philipplackner

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
steve1rm: @MKiperszmid I named it errorData as I will create a workDataOf to show the error if the upload failed. What do you think this should be called?

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
steve1rm: @MKiperszmid Thanks for confirming this

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
steve1rm: @MKiperszmid I am sure that @philipplackner said we get back the key and url in the response. And these are the keys we use when we want to delete the photos from the BE. Maybe @philipplackner can confirm what we need to do with this response.

data/src/main/java/me/androidbox/data/local/converter/EventConverter.kt
```diff
@@ -0,0 +1,25 @@
+package me.androidbox.data.local.converter
+
+import androidx.room.TypeConverter
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
+import me.androidbox.data.remote.model.request.EventCreateRequestDto
+import javax.inject.Inject
+
+class EventConverter @Inject constructor(){
```
philipplackner, line 9: If you have an @Inject constructor, you don't need a @Provides function for this class

data/src/main/java/me/androidbox/data/remote/network/EndPointConstant.kt
philipplackner: There isn't even a real use case for having constants for URLs at all. They're just used at one place and a constant also doesn't add readability

data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
philipplackner: When using the network connected constraints for WorkManager, that's handled. Ideally together with a backoff strategy

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
philipplackner: Yes correct @steve1rm . You need these keys and the URLs are also only known after the API responds

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
philipplackner: Yes, but there's a typo, you called it `errorDate`

data/src/main/java/me/androidbox/data/worker_manager/util/CreatePhotoMultipart.kt
philipplackner: Also it rather sounds like a function and not an interface

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
philipplackner: The only thing I can think of is if the device is out of storage

philipplackner (approved):

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
steve1rm: Should be errorData

data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
```diff
@@ -0,0 +1,55 @@
...
+            IS_EDIT_MODE to isEditMode,
+            EVENT to eventRequestJson,
+            EVENT_PHOTOS to event.photos.toTypedArray())
+
+        val uploadWorkerRequest  = OneTimeWorkRequestBuilder<UploadEventWorker>()
+            .setConstraints(Constraints(Constraints.Builder().setRequiredNetworkType(NetworkType.CONNECTED).build()))
+            .setBackoffCriteria(
+                BackoffPolicy.EXPONENTIAL,
+                WorkRequest.MIN_BACKOFF_MILLIS,
+                TimeUnit.SECONDS)
```
steve1rm: @philipplackner  @MKiperszmid Is this the best way to handle a backoffcriteria?

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
```diff
@@ -0,0 +1,106 @@
...
+
+        if(!eventRequestJson.isNullOrEmpty()) {
+            val listOfPhotoMultiPart = listOfPhoto?.let {
+                /* Create the multipart for the list of photos */
+                createPhotoMultipart.createMultipartPhotos(listOfPhoto.toList())
+            } ?: listOf()
+
+            if(runAttemptCount > RETRY_COUNT) {
+                Result.failure()
+            }
```
steve1rm, line 50: @philipplackner @MKiperszmid Adds a RETRY_COUNT

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
```diff
@@ -0,0 +1,106 @@
...
+
+                        Result.success(outputData)
+                    }
+                    else {
+                        Result.success()
+                    }
+                },
+                onFailure = { _ ->
+                    Result.retry()
+                }
```
steve1rm, line 72: @philipplackner @MKiperszmid Adds retry in failure.

data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
philipplackner: Yes, just that you assign a milliseconds value but specify it as seconds

philipplackner (approved):

PR #29, opened 2023-03-29
[27] visitor particpate component

steve1rm:
# Pull Request template

###  What does this PR solve?

- Add visitor item component
- Adds event screen state and event

visitorItem


component/src/main/java/me/androidbox/component/event/VisitorItem.kt
```diff
@@ -0,0 +1,104 @@
...
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import me.androidbox.component.R
+import me.androidbox.component.ui.theme.*
+
+@Composable
+fun VisitorItem(
+    visitorInitials: String,
```
steve1rm, line 27: The initials for the visitor will be extracted outside this VisitorItem and passed in. So it might seen I could do this inside this VisitorItem as have full name of the user.

philipplackner: I'd leave it outside, so you're more flexible and could also pass in custom initials that don't require parsing a full name before

component/src/main/java/me/androidbox/component/event/VisitorItem.kt
```diff
@@ -0,0 +1,104 @@
...
+        verticalAlignment = Alignment.CenterVertically) {
+        Text(
+            modifier = Modifier
+                .padding(12.dp)
+                .drawBehind {
+                    drawCircle(
+                        color = Color.LightGray,
+                        radius = this.size.maxDimension
+                    )
+                },
```
philipplackner, line 45: No need for a canvas here, you can simply clip the content:
```
Modifier.clip(CircleShape).background(Color.LightGray)
```

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
```diff
@@ -6,5 +6,6 @@ import me.androidbox.component.agenda.VisitorType
 
 data class EventScreenState(
     val listOfPhotoUri: List<Uri> = mutableStateListOf<Uri>(),
-    val selectedVisitorType: VisitorType = VisitorType.ALL
+    val selectedVisitorType: VisitorType = VisitorType.ALL,
+    val selectedVisitor: String = "",
```
philipplackner: Isn't there more to a visitor than just the name (assuming this string refers to the name which isn't clear)? If this refers to the name, I'd make it refer to the ID since names can appear multiple times. And if it refers to the ID, I'd also name it that way (`selectedVisitorId`)

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -35,6 +37,11 @@ class EventViewModel @Inject constructor(
                     selectedVisitorType = eventScreenEvent.visitorType
                 )
             }
+            is EventScreenEvent.OnDeleteVisitor -> {
+                _eventScreenState.value = eventScreenState.value.copy(
+                    selectedVisitor = eventScreenEvent.visitorName
```
philipplackner: Okay, so apparently it refers to the name, then see my comment above 👍

philipplackner (approved):

component/src/main/java/me/androidbox/component/event/VisitorInfo.kt
```diff
@@ -0,0 +1,6 @@
+package me.androidbox.component.event
+
+data class VisitorInfo(
+    val visitorName: String = "",
+    val visitorId: String = ""
+)
```
philipplackner: Good, just that I wouldn't use default parameters for non-nullable strings this way. If you want to indicate that a value doesn't exist, make it nullable - that's in the end what nullables are used for. In this case I would keep the strings as non-null and not initialize them, but rather work with a `VisitorInfo?`. Either a visitor exists or not, but if they exist they certainly have a name and an ID

philipplackner (approved):

component/src/main/java/me/androidbox/component/event/VisitorInfo.kt
steve1rm: @philipplackner Updated the VisitorInfo and removed default arguments.

component/src/main/java/me/androidbox/component/event/VisitorItem.kt
steve1rm: @philipplackner Trying this here:
```
        Text(
            modifier = Modifier
                .clip(CircleShape)
                .background(Color.LightGray),
            text = visitorInitials,
            fontSize = 12.sp,
            color = MaterialTheme.colorScheme.visitorInitialsFontColor,
            fontWeight = FontWeight.SemiBold
        )
```
Screenshot from 2023-03-29 20-37-12

PR #30, opened 2023-03-29
[28] agender action delete join leave

steve1rm:
# Pull Request template

###  What does this PR solve?

Adds the agenda actions that could be delete, join, leave. For events, tasks, and reminders

agenda-actions


component/src/main/java/me/androidbox/component/agenda/AgendaAction.kt
```diff
@@ -0,0 +1,102 @@
...
+        /* Reminders and Tasks have a top divider */
+        if(agendaActionType.showDivider) {
+            Divider(
+                modifier
+                    .fillMaxWidth()
+                    .padding(horizontal = 16.dp)
+                    .background(
+                        color = MaterialTheme.colorScheme.divider,
+                        shape = RoundedCornerShape(8.dp)
+                    ),
```
philipplackner: A Divider already has a `color` attribute for this which I'd recommend to use

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -35,6 +35,11 @@ class EventViewModel @Inject constructor(
                     selectedVisitorType = eventScreenEvent.visitorType
                 )
             }
+            is EventScreenEvent.OnSelectedAgendaAction -> {
+                _eventScreenState.value = eventScreenState.value.copy(
+                    selectedAgendaActionType = eventScreenEvent.agendaActionType
+                )
+            }
```
philipplackner: Have I commented on using `_stateFlow.update` before? In case not:

In general, I'd always use `_stateFlow.update { it.copy(...) }` to avoid race conditions. This is how it could happen:

*StateFlow holds value A*
- Thread 1 calls `copy()` on StateFlow with value A
- Before Thread 1 assigns its changed copy, Thread 2 updates StateFlow with value B
- Now, Thread 1 assigns its changed copy, which is still based on value A though, but the latest value is already value B. Therefore, value B would never be considered here.

Not always an issue, but I worked on highly stateful apps with frequently changing states, in which this really bit me, since such bugs are super hard to find.
Using `_state.value = _state.value.copy(...)` is fine if you always call this from a single thread. But in ViewModels, you're very likely to not do that as soon as you'd change the state inside a `viewModelScope`.

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
steve1rm: @philipplackner Updated to use `update` on the stateflow

PR #34, opened 2023-03-30
[32] agenda idetail topbar for saving and editing

steve1rm:
# Pull Request template

###  What does this PR solve?
Adds Agenda Details topbar

topbar-agendadetails


component/src/main/java/me/androidbox/component/agenda/AgendaDetailTopBar.kt
```diff
@@ -0,0 +1,124 @@
...
+    onSaveClicked: () -> Unit
+) {
+    Row(modifier = modifier,
+        verticalAlignment = Alignment.CenterVertically,
+    horizontalArrangement = Arrangement.SpaceBetween) {
+        IconButton(onClick = {
+            onCloseClicked()
+        }) {
+            Icon(painterResource(id = R.drawable.close_white),
+                tint = Color.White,
```
steve1rm, line 42: @philipplackner @MKiperszmid Just wondering if this icon is white, why do we need to set the tint color. Without it the icon color is black

component/src/main/java/me/androidbox/component/agenda/AgendaDetailTopBar.kt
```diff
@@ -0,0 +1,124 @@
...
+import me.androidbox.component.ui.theme.topbarFontColor
+
+@Composable
+fun AgendaDetailTopBar(
+    editModeType: EditModeType,
+    displayDate: String,
+    modifier: Modifier = Modifier,
+    onCloseClicked: () -> Unit,
+    onEditClicked: () -> Unit,
+    onSaveClicked: () -> Unit
```
MKiperszmid: These callbacks should be before the modifier, or else you can only access them via specifying the parameter

component/src/main/java/me/androidbox/component/agenda/AgendaDetailTopBar.kt
MKiperszmid: Sometimes when importing icons on the drawable folder, it defaults the color to black
Maybe that's the case?

component/src/main/java/me/androidbox/component/agenda/AgendaDetailTopBar.kt
```diff
@@ -0,0 +1,124 @@
...
+                    onEditClicked()
+                }) {
+                    Icon(painterResource(id = editModeType.editIcon),
+                        tint = Color.White,
+                        contentDescription = null)
+                }
+            }
+
+            EditModeType.SAVE -> {
+                Text(
```
MKiperszmid: When a Text is clickable, you can use the "TextButton" composable

component/src/main/java/me/androidbox/component/agenda/AgendaDetailTopBar.kt
```diff
@@ -0,0 +1,124 @@
...
+    }
+}
+
+enum class EditModeType(
+    @DrawableRes val editIcon: Int = R.drawable.pencil,
+    @StringRes val saveRes: Int = R.string.save
+) {
+    IS_EDIT(editIcon = R.drawable.pencil),
+    SAVE(saveRes = R.string.save)
+}
```
MKiperszmid: Here since depending on the type of "EditMode" you show either an icon or a text, you could use a Sealed Class

component/src/main/res/drawable/close_white.xml
```diff
@@ -0,0 +1,12 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="18dp"
+    android:height="18dp"
+    android:viewportWidth="18"
+    android:viewportHeight="18">
+  <path
+      android:pathData="M17,17L1,1M17,1L1,17"
+      android:strokeWidth="2"
+      android:fillColor="#00000000"
```
MKiperszmid, line 9: Here

The fill color is black

So maybe that's the issue you were getting about the icon being black?

domain/src/main/java/me/androidbox/domain/authentication/model/Event.kt
```diff
@@ -1,14 +1,14 @@
...
+    val id: String = "",
+    val title: String = "",
+    val description: String = "",
+    val startDateTime: Long = 0L,
+    val endDateTime: Long = 0L,
+    val remindAt: Long = 0L,
+    val eventCreatorId: String = "",
+    val isUserEventCreator: Boolean = false,
+    val attendees: String = "",
+    val photos: String = "",
```
MKiperszmid: What's the use of adding default parameters here?

You would never need an empty Event

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -63,6 +63,22 @@ class EventViewModel @Inject constructor(
                     )
                 }
             }
+            is EventScreenEvent.OnSaveEventDetails -> {
+                _eventScreenState.update { eventScreenState ->
+                    eventScreenState.copy(
+                        saveEventDetails = eventScreenEvent.event
```
MKiperszmid: Wouldn't you have this Event on the state? Why are you passing it through an Event (EventScreenEvent.OnSaveEventDetails) ?

MKiperszmid (approved):

domain/src/main/java/me/androidbox/domain/authentication/model/Event.kt
steve1rm: @MKiperszmid the reason I added default parameters because I set this state in the EventScreenState. What should I be doing?

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
steve1rm: @MKiperszmid not sure I understand. This Event will contain the event information that has been set by the user. Then will be saved to the room dB. What should I be doing here?

domain/src/main/java/me/androidbox/domain/authentication/model/Event.kt
MKiperszmid: One solution is not use an Event in the state, simply create the Event when you have all the info

So on "saveEvent()" you can create the event there with all the info from the state

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
MKiperszmid: image
Yeah but you have all the info to create the Event in the State

So why pass the Event over the Screen to the ViewModel, if the viewmodel already has all the info?

component/src/main/java/me/androidbox/component/agenda/AgendaDetailTopBar.kt
steve1rm: @MKiperszmid updated

component/src/main/java/me/androidbox/component/agenda/AgendaDetailTopBar.kt
steve1rm: @MKiperszmid Updated to use TextButton

domain/src/main/java/me/androidbox/domain/authentication/model/Event.kt
steve1rm: @MKiperszmid Removed the default parameters

component/src/main/java/me/androidbox/component/agenda/AgendaDetailTopBar.kt
steve1rm: @MKiperszmid Updated to use a sealed interface

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -63,6 +63,27 @@ class EventViewModel @Inject constructor(
...
+                /* Then insert this event into the room db
+                val Event = Event(
+                    id = UUID.randomUUID().toString(),
+                    description = eventScreenState.value.saveEditOrDescriptionContent,
+                    photos = eventScreenState.value.listOfPhotoUri,
+                    isUserEventCreator = true,
+                    ...
+                )
+                   insertEventDetails(event)
+                 */
```
steve1rm, line 79: So it will be something like the above.
1) Extract the information from the EventScreenState i.e. title, description, photo urls, id, visitors, etc.
2) Create the Event object
3) Insert that event into the DB

component/src/main/java/me/androidbox/component/agenda/AgendaDetailTopBar.kt
philipplackner: Yes I think if your device is in light mode, it defaults to black, so it looks good on a light surface

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
philipplackner: Looks good

philipplackner (approved):

PR #36, opened 2023-04-02
[34] date time event duration for event

steve1rm:
# Pull Request template

###  What does this PR solve?

- Adds the time and date for the event duration
- Updates the state using the event
- Only uses 2 button to update both the time and date for either start or end

https://user-images.githubusercontent.com/14260802/229340829-3929e42e-c471-4011-a2ba-4582882e9a43.mp4

### DIFF
--- a/presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
+++ b/presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
@@ -148,8 +148,8 @@ fun EventScreen(
                 eventScreenEvent(EventScreenEvent.OnStartDateDuration(zonedDateTime))
             }
             else {
-                eventScreenEvent(EventScreenEvent.OnStartTimeDuration(zonedDateTime))
-                eventScreenEvent(EventScreenEvent.OnStartDateDuration(zonedDateTime))
+                eventScreenEvent(EventScreenEvent.OnEndTimeDuration(zonedDateTime))
+                eventScreenEvent(EventScreenEvent.OnEndDateDuration(zonedDateTime))
             }
         }
     )
--- a/presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
+++ b/presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
@@ -16,6 +16,8 @@ import me.androidbox.component.general.PhotoPicker
 import me.androidbox.component.ui.theme.BusbyTaskyTheme
 import me.androidbox.component.ui.theme.backgroundBackColor
 import me.androidbox.component.ui.theme.backgroundWhiteColor
+import java.time.format.DateTimeFormatter
+import java.util.*
 
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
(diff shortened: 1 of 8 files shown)


presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -111,6 +133,26 @@ fun EventScreen(
...
+            if(isStartDateTime) {
+                eventScreenEvent(EventScreenEvent.OnStartTimeDuration(zonedDateTime))
+                eventScreenEvent(EventScreenEvent.OnStartDateDuration(zonedDateTime))
+            }
+            else {
+                eventScreenEvent(EventScreenEvent.OnEndTimeDuration(zonedDateTime))
+                eventScreenEvent(EventScreenEvent.OnEndDateDuration(zonedDateTime))
+            }
+        }
+    )
```
steve1rm, line 154: @MKiperszmid @philipplackner In this DateTimeDialog I am setting both the time and the date for the start and end duration. So when the dialog opens the user will set the date and time in one go.

I have added a boolean flag to indicate which should be set the start or the end duration.

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -26,6 +35,9 @@ fun EventScreen(
     onEditDescriptionClicked: (description: String) -> Unit,
     modifier: Modifier = Modifier) {
 
+    val calendarStateTimeDate = rememberUseCaseState()
+    var isStartDateTime = false
```
philipplackner: This is a side-effect (something that's not Compose related, but is inside of a composable context). Every single time `EventScreen` recomposes, this `isStartDateTime` will be false again.

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -78,14 +90,24 @@ fun EventScreen(
...
+                        endDate = eventScreenState.endDateDuration.formatDateTime(DATE_PATTERN),
+                        modifier = Modifier.fillMaxWidth(),
+                        onStartDurationClicked = {
+                            isStartDateTime = true
+                            calendarStateTimeDate.show()
+                        },
+                        onEndDurationClicked = {
+                            isStartDateTime = false
+                            calendarStateTimeDate.show()
+                        }
```
philipplackner: I'd completely remove the `isStartDateTime` boolean for several reasons:
1. As mentioned above, it's a side-effect and will lead to undefined behavior
2. Whether your date picker dialog is shown or not (and also which one) is considered state. That boolean should therefore go in your screen state class and the ViewModel should toggle it. That way it'll also survive config changes such as toggling the device theme.

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
```diff
@@ -5,12 +5,18 @@ import androidx.compose.runtime.mutableStateListOf
...
     val selectedVisitorType: VisitorType = VisitorType.ALL,
     val selectedVisitor: VisitorInfo? = null,
     val eventTitle: String = "New Event",
-    val eventDescription: String = "New Description",
+    val eventDescription: String = "Description",
     val selectedAgendaActionType: AgendaActionType = AgendaActionType.DELETE_EVENT,
+    val startTimeDuration: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault()),
+    val endTimeDuration: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault()),
+    val startDateDuration: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault()),
+    val endDateDuration: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault())
```
philipplackner: All these are not really durations, but rather a fixed time/date

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
steve1rm: @philipplackner Updated and moved this to the EventScreenState and update using an event in the EventViewModel

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
steve1rm: Updated to startTime, startDate, endTime, and endDate

PR #38, opened 2023-04-03
[36] alarm manager notification

steve1rm:
# Pull Request template

###  What does this PR solve?

- Adds the alarm manager for setting reminders for events, tasks, and reminders. Currently only implemented for Events. But tasks and reminders will be coming soon.  But the current implementation should work for all as the ID for those is used to set the alarm.

1) User creates a new event and set the time they want to be reminded
2) When user saves the event this event will be inserted into the DB
3) After inserting an alarm will be scheduled
4) An intent is created with the ID, title, description, agenda type, and remindAt extras
5) In the onReceive the extras are extracted and used to create the notification.

**What is not implemented yet:**

- Clicking on the notification and navigating to the Event Detail Screen.
- Resetting alarms after a reboot.


presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -112,6 +113,8 @@ fun EventScreen(
 
                     Spacer(modifier = modifier.height(26.dp))
 
+                    val context: Context = LocalContext.current
```
steve1rm: @philipplackner @MKiperszmid Is it ok to have this context declared here. As I need to pass it to the `setupAlarmReminderItem`  to obtain the string from the resource id?

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -187,6 +187,34 @@ fun EventScreen(
...
+                dateTime = startTime.minusMinutes(startTime.toEpochSecond()),
+                message = context.getString(eventScreenState.alarmReminderText)
+            )
+        }
+        4 -> {
+            val startDate = eventScreenState.startDate
+
+            AlarmItem(
+                dateTime = startDate.minusDays(startDate.toEpochSecond()),
+                message = context.getString(eventScreenState.alarmReminderText)
```
steve1rm: @philipplackner @MKiperszmid The context is passed in here so I can use getString. This was to avoid the `Functions which invoke @Composable functions must be marked with the @Composable annotation`

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
```diff
@@ -23,6 +23,6 @@ data class EventScreenState(
     val endDate: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault()),
     val isStartDateTime: Boolean = false,
     val shouldOpenDropdown: Boolean = false,
-    val alarmItem: AlarmItem = AlarmItem(ZonedDateTime.now(), null),
+    val alarmItem: AlarmItem = AlarmItem(ZonedDateTime.now(), ""),
```
steve1rm: @philipplackner @MKiperszmid I am just wondering if the `AlarmItem` should be null or not.

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -129,12 +132,9 @@ fun EventScreen(
...
 
-                            /** TODO Using mock data to set the Alarm Monitor, here use the eventScreenState start time date */
-                            val alarmItem = AlarmItem(
-                                ZonedDateTime.now().plusSeconds(30L),
-                                "THIS IS THE MESSAGE FROM THE ALARM REMINDER"
-                            )
-                            eventScreenEvent(EventScreenEvent.OnScheduleAlarmReminder(alarmItem))
+                            setupAlarmReminderItem(eventScreenState = eventScreenState, item = item, context)?.let { alarmItem ->
+                                eventScreenEvent(EventScreenEvent.OnScheduleAlarmReminder(alarmItem))
+                            }
```
steve1rm: @philipplackner @MKiperszmid If something goes wrong then the `setupAlarmReminderItem` will return a nullable AlarmItem. Is this ok?

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
```diff
@@ -1,14 +1,47 @@
...
+            val pendingIntent = PendingIntent.getActivity(context, 1, notificationIntent, PendingIntent.FLAG_IMMUTABLE)
+            val notification = NotificationCompat.Builder(context, CHANNEL_ID)
+                .setContentTitle(context.getString(R.string.notification_title))
+                .setSmallIcon(R.drawable.bell)
+                .setContentText(message)
+                .setAutoCancel(true)
+                .setContentIntent(pendingIntent)
+                .build()
+
+            notificationManager.notify(1, notification)
```
MKiperszmid: What if a user has it set so it has a reminder, an event and 1 task at the same time?

With .notify(1  you are setting the same id always, so the notification will get overridden
You could use something like:

1st - You should also pass the event/reminder/task id
2nd - Use notify with the id being the hashcode of the item (Ej: reminder.id.hashcode())

This makes it so each notification is unique to an item´s id

MKiperszmid: Also, when a notification get´s clicked, you should send the user to the event/reminder/task detail screen

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
```diff
@@ -1,14 +1,47 @@
...
+                    NotificationManager.IMPORTANCE_HIGH
+                )
+
+                notificationManager.createNotificationChannel(channel)
+            }
+
+            val notificationIntent = Intent(context, AlarmReceiver::class.java)
+            val pendingIntent = PendingIntent.getActivity(context, 1, notificationIntent, PendingIntent.FLAG_IMMUTABLE)
+            val notification = NotificationCompat.Builder(context, CHANNEL_ID)
+                .setContentTitle(context.getString(R.string.notification_title))
```
MKiperszmid: The notification should display as title, the title of the event/task/reminder
And as contenttext, the description

domain/src/main/java/me/androidbox/domain/alarm_manager/AlarmItem.kt
```diff
@@ -2,12 +2,7 @@ package me.androidbox.domain.alarm_manager
 
 import java.time.ZonedDateTime
 
-
 data class AlarmItem(
```
MKiperszmid, line 5: What´s the use of this item?

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
MKiperszmid: It´s fine to get the context so you can use stringResources

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
MKiperszmid: I'm not understanding the use of this

You should just save a ZonedDateTime, nothing else

Then, once the alarm get's triggered, you should get the event/task/reminder, you should get all info from the repository (db) based on the id

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
MKiperszmid: Yea that's fine

But regarding this code. All this logic shouldn't happen on the UI

You should calculate the time on the ViewModel at least

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
steve1rm: @MKiperszmid I understand that we send reminders to each agenda item (event, tasks, and reminders). And the alarm manager gets this information from room DB.

1) User creates a new event with the start, end, time, date, reminder time, etc, and saves that to the DB
2) This will be uploaded to the API as well.
3) This is the bit I am not sure about, if the alarm manager gets the information from the DB, each time a reminder is updated or created. How does it know it needs to do that? 

Could it be after a insert or update has happened the alarm manager has to query the DB.

domain/src/main/java/me/androidbox/domain/alarm_manager/AlarmItem.kt
steve1rm: @MKiperszmid I was following @philipplackner tutorial on how to use Alarm Manager, as the time I have never used Alarm Manager. But I guess I can remove this if its not needed.

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
steve1rm: @MKiperszmid if we save the just the ZoneDateTime and nothing else. For example the alarm is set to go off at 3pm. It then triggers a query to the DB to get the title, description and show the notification. However, it will also need the ID of the event to know which event it needs to query for. Am I missing something here?

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
philipplackner: Yes, the reason we do that in the ViewModel is that we then have a single testable source of truth for our state. If another UI component starts depending on it and you do it individually for all components on the UI every time, it can quickly cause bugs when requirements change

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
philipplackner: There's no need for the alarm to query the DB. The flow is like the following:
1. User saves agenda item with reminder type set to something like 30min before
2. That translates for example to a remindAt timestamp which is 30min before the agenda item time
3. You use AlarmManager to schedule an action with that remindAt timestamp. That requires passing an Intent to which you can pass all the info that is needed on the notification (title, description, etc. - in the end your `AlarmItem`). From that on it will stay scheduled
4. At the scheduled timestamp, AlarmManager will send a broadcast to your app which you can receive in a BroadcastReceiver
5. In that receiver, you get the passed data from step 3 and you can show the notification with that
6. When the device reboots, all alarms are deleted again. Therefore, after device reboot, you need to go through all future agenda items and reset their alarms (as I mentioned in my earlier message)

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
philipplackner: This does not seem to be the place where you want to schedule alarms. As far as I understand, this schedules a new alarm every time the user selects a different reminder option, but what if they then decide to not save the item? You should be able to do all the alarm logic in the ViewModel itself, the UI doesn't need to trigger it.

philipplackner (approved):

domain/src/main/java/me/androidbox/domain/alarm_manager/AlarmItem.kt
philipplackner: A similar item like that will be needed. In the end you need to tell AlarmManager everything the notification needs to know, e.g. when it's shown, the title, the description, the type of agenda item, the time for the item, etc.

Edit: It depends actually. If you get the items from your Room DB in the BroadcastReceiver, you don't need this AlarmItem class

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
philipplackner: Simply see AlarmManager as something that will notify your app at a given timestamp and not more. You can then receive and respond to this trigger by using a `BroadcastReceiver`. This will give you an Intent from which you can extract extras, like the agenda item ID the alarm was triggered for. You then take this ID and load the item from the DB to show the notification in the broadcast receiver.
There's no need to listen to insertions, etc. But you will need to update the alarm if the user updates the reminder type or the agenda item time

philipplackner (approved):

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
```diff
@@ -1,15 +1,75 @@
...
+import me.androidbox.data.alarm_manager.AlarmSchedulerImp.Companion.EXTRA_TITLE
 
 class AlarmReceiver : BroadcastReceiver() {
 
     override fun onReceive(context: Context?, intent: Intent?) {
-        intent?.getStringExtra("EXTRA_MESSAGE")?.let { message ->
-            /* TODO Add Notification here */
-            println("Alarm triggered: $message")
+        /* TODO Not sure if intent and context can be null but I guess is ok to cast it anyway */
+        if(intent != null && context != null) {
```
steve1rm, line 21: @MKiperszmid @philipplackner Do we need to check for null context and intent. Would they really be null?

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
```diff
@@ -1,15 +1,75 @@
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val notificationManager = getNotificationManager(context)
+            val notificationChannel = NotificationChannel(
+                agendaId,
+                agendaName,
+                NotificationManager.IMPORTANCE_HIGH
+            )
+
+            notificationManager.createNotificationChannel(notificationChannel)
         }
```
steve1rm, line 48: @MKiperszmid @philipplackner Is it ok to use the agendaId for the channelId, which could be the eventId, taskId, or reminderId for what alarm is be scheduled for?

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
MKiperszmid: I never had a null intent and context, but I prefer checking them just in case


Here to make it a bit more clearer I like to do:

if(intent == null || context == null)  return


So you just add that at the top, and done
You don't need to put everything inside an if

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
MKiperszmid: Nono this is bad
Since you'll create a new channel for each agenda item, and that's a big no-no

You could have either
1 channel to display all agendaitem notifications
Or 3 channels, one for each type of notification

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
```diff
@@ -1,15 +1,75 @@
...
     }
+
+    private fun createAndShowNotification(context: Context, agendaId: String, title: String, description: String) {
+        val notificationManager = getNotificationManager(context)
+        val notificationIntent = Intent(context, AlarmReceiver::class.java)
+        val pendingIntent = PendingIntent.getActivity(
+            context,
+            agendaId.hashCode(),
+            notificationIntent,
```
MKiperszmid: Notificationintent is set to AlarmReceiver class

Won't clicking the notification do nothing? Or in fact, crash?

data/src/main/java/me/androidbox/data/alarm_manager/AlarmSchedulerImp.kt
```diff
@@ -13,30 +13,43 @@ class AlarmSchedulerImp @Inject constructor(
     @ApplicationContext private val context: Context,
```
MKiperszmid, line 13: No need to add @ApplicationContext , if you already have it on the Hilt Module

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
```diff
@@ -23,6 +23,6 @@ data class EventScreenState(
     val endDate: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault()),
     val isStartDateTime: Boolean = false,
     val shouldOpenDropdown: Boolean = false,
-    val alarmItem: AlarmItem = AlarmItem(ZonedDateTime.now(), null),
-    @StringRes val alarmReminderText: Int = me.androidbox.component.R.string.ten_minutes_before,
+    @StringRes val alarmReminderText: Int = R.string.ten_minutes_before,
```
MKiperszmid: Do you really need this?

One option could be to have the AlarmReminderItem receive the stringRes as parameter
So you don't do it here in the State

Mostly so the Enum takes care of everything about itself

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -123,31 +116,40 @@ class EventViewModel @Inject constructor(
...
+        val event = Event(
+            id = UUID.randomUUID().toString(),
+            title = eventScreenState.value.eventTitle,
+            description = eventScreenState.value.eventDescription,
+            startDateTime = startDateTime.toEpochSecond(),
+            endDateTime = endDateTime.toEpochSecond(),
+            remindAt = remindAt.toEpochSecond(),
+            eventCreatorId = preferenceRepository.retrieveCurrentUserOrNull()?.userId ?: "",
+            isUserEventCreator = false,
+            attendees = "attendee 1", /* TODO Add attendees here */
```
MKiperszmid, line 143: I know this is mocked data, but remember attendees is a list, so it's better to have a list

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
steve1rm: @MKiperszmid @philipplackner  Should this be my `HomeActivity::class.java`

As I haven't completed the deeplinks yet. I just want the notification to do nothing if the user clicks it. Will that work

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
steve1rm: @MKiperszmid 
I have used the following for the channel name, which will appear in the app settings on the device:
`val channelName = "Notifications for $agendaType"`

And for the channel ID I am using the `agendaType` (event, task, reminder)

So would this create a channel for notifications for each agenda type, event, task, or reminder

Would this be correct now?

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
MKiperszmid: This pending intent isn´t used for the time being, so can be removed and added later once you add deeplinks

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
MKiperszmid: Yea! That´s good

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
philipplackner: I think technically it could happen if whatever sends the broadcast (here AlarmManager) doesn't attach anything, but for AlarmManager you can be pretty sure that doesn't happen

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
```diff
@@ -1,15 +1,69 @@
...
+        /* TODO Not sure if intent and context can be null but I guess is ok to cast it anyway */
+        if(intent != null && context != null) {
+            intent.extras?.let { bundle ->
+                val agendaId = bundle.getString(EXTRA_ID) ?: ""
+                val title = bundle.getString(EXTRA_TITLE) ?: "Title"
+                val description = bundle.getString(EXTRA_DESCRIPTION) ?: "Description"
+                val agendaType = bundle.getString(EXTRA_AGENDA_TYPE) ?: "Agenda"
+
+                if (agendaId.isNotBlank()) {
+                    createNotificationChannel(context, agendaType)
```
philipplackner, line 29: I'd create the notification channel in `onCreate` of your application class, so directly when your app launches. Channels are used so that users can decide in the app settings whether they want to receive a certain type of notifications or not. If you create the channel just before showing a notification, they won't be able to change that setting before

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
philipplackner: Yes, one channel per notification type is great

data/src/main/java/me/androidbox/data/alarm_manager/AlarmSchedulerImp.kt
philipplackner: But if you don't provide this class in the Hilt module, you do need it here

data/src/main/java/me/androidbox/data/alarm_manager/AlarmSchedulerImp.kt
```diff
@@ -13,30 +13,43 @@ class AlarmSchedulerImp @Inject constructor(
...
-        }
+    override fun scheduleAlarmReminder(alarmItem: AlarmItem) {
+        /** Only set the alarm when the remindAt time is less than the current time */
+        if (System.currentTimeMillis() < alarmItem.remindAt) {
+            val alarmIntent = Intent(context, AlarmReceiver::class.java).apply {
+                putExtra(EXTRA_ID, alarmItem.agendaId)
+                putExtra(EXTRA_TITLE, alarmItem.title)
+                putExtra(EXTRA_DESCRIPTION, alarmItem.description)
+                putExtra(EXTRA_AGENDA_TYPE, alarmItem.agendaType.name)
+            }
```
philipplackner, line 33: You could also think about passing the whole `AlarmItem` as an extra (and making it a `Parcelable` first). That way if you end up needing more info in future to create a notification, you just need to extend the `AlarmItem` class, but not add a new extra here every time.

data/src/main/java/me/androidbox/data/alarm_manager/AlarmSchedulerImp.kt
```diff
@@ -13,30 +13,43 @@ class AlarmSchedulerImp @Inject constructor(
...
-                intent,
-                PendingIntent.FLAG_UPDATE_CURRENT
+            alarmManager.setExact(
+                AlarmManager.RTC_WAKEUP,
+                alarmItem.remindAt,
+                PendingIntent.getBroadcast(
+                    context,
+                    alarmItem.agendaId.hashCode(),
+                    alarmIntent,
+                    PendingIntent.FLAG_UPDATE_CURRENT
```
philipplackner, line 42: For API 31+ you should also add `PendingIntent.FLAG_IMMUTABLE` which is required since then. Otherwise I think the app will crash on these devices

domain/src/main/java/me/androidbox/domain/alarm_manager/AlarmItem.kt
```diff
@@ -1,13 +1,27 @@
...
+    val remindAt: Long,
+    val agendaType: AgendaType
+)
 
+fun Event.toAlarmItem(agendaType: AgendaType): AlarmItem {
+    return AlarmItem(
+        agendaId = this.id,
+        title = this.title,
+        description = this.description,
+        remindAt = this.remindAt * 1000,
```
philipplackner, line 18: Why not have one consistent scale for timestamps? Nothing tells me here that `remindAt` for the event is in seconds and for the alarm item in milliseconds. Things like these will 99% lead to a bug especially when working together with other devs.
So I'd just default all timestamps in the app to be in milliseconds or seconds

presentation/src/main/java/me/androidbox/presentation/alarm_manager/AlarmReminderProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+                startTime.minusHours(1L)
+            }
+            AlarmReminderItem.SIX_HOUR -> {
+                startTime.minusHours(6L)
+            }
+            AlarmReminderItem.ONE_DAY -> {
+                startDate.minusDays(1L)
+            }
+        }
+    }
```
philipplackner, line 32: This could also simply be a member function of `AlarmReminderItem`:
```
enum class AlarmReminderItem {
    TEN_MINUTES, ..., ONE_DAY;

    fun getRemindAt(startTime: ZonedDateTime): ZonedDateTime {
        // Your when expression here
    }
}

// Then called like this:
val remindAt = reminderItem.getRemindAt(startTime)
```
I find this a bit more readable than a separate class which you would not need in that case.

presentation/src/main/java/me/androidbox/presentation/alarm_manager/AlarmReminderProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+import me.androidbox.component.agenda.AlarmReminderItem
+import java.time.ZoneId
+import java.time.ZonedDateTime
+
+object AlarmReminderProvider {
+
+    fun getRemindAt(
+        alarmReminderItem: AlarmReminderItem,
+        startTime: ZonedDateTime,
+        startDate: ZonedDateTime
```
philipplackner, line 12: `ZonedDateTime` already contains both date and time as the name says, so only one value is needed here

PR #39, opened 2023-04-05
[37] notification update

steve1rm:
# Pull Request template

###  What does this PR solve?

- Creates the notification channel in the onCreate of the application class
- Changes the name to remindAtInMilliSecs
- Adds FLAG_IMMUATABLE
- Uses the DateTime instead of using the date and time separately

### DIFF
--- a/domain/src/main/java/me/androidbox/domain/alarm_manager/AlarmItem.kt
+++ b/domain/src/main/java/me/androidbox/domain/alarm_manager/AlarmItem.kt
@@ -6,7 +6,7 @@ data class AlarmItem(
     val agendaId: String,
     val title: String,
     val description: String,
-    val remindAt: Long,
+    val remindAtInMilliSecs: Long,
     val agendaType: AgendaType
 )
 
@@ -15,7 +15,7 @@ fun Event.toAlarmItem(agendaType: AgendaType): AlarmItem {
         agendaId = this.id,
         title = this.title,
         description = this.description,
-        remindAt = this.remindAt * 1000,
+        remindAtInMilliSecs = this.remindAt * 1000,
         agendaType = agendaType
     )
 }
--- a/domain/src/main/java/me/androidbox/domain/alarm_manager/AlarmItem.kt
+++ b/domain/src/main/java/me/androidbox/domain/alarm_manager/AlarmItem.kt
@@ -6,7 +6,7 @@ data class AlarmItem(
     val agendaId: String,
     val title: String,
     val description: String,
-    val remindAtInMilliSecs: Long,
+    val remindAt: Long,
     val agendaType: AgendaType
 )
 
@@ -15,7 +15,7 @@ fun Event.toAlarmItem(agendaType: AgendaType): AlarmItem {
         agendaId = this.id,
         title = this.title,
         description = this.description,
-        remindAtInMilliSecs = this.remindAt * 1000,
+        remindAt = this.remindAt,
         agendaType = agendaType
     )
 }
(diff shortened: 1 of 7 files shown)


presentation/src/main/java/me/androidbox/presentation/BusbyTaskyApplication.kt
```diff
@@ -1,7 +1,45 @@
...
+        return this.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+    }
+
+    private fun generateChannels(): Map<String, String> {
+        return mapOf(
+            "Notifications for EVENTS" to "event",
+            "Notifications for TASKS" to "task",
+            "Notifications for REMINDERS" to "reminders"
+        )
+    }
```
steve1rm, line 44: @philipplackner Is this the best way to create 3 separate channels

domain/src/main/java/me/androidbox/domain/alarm_manager/AlarmItem.kt
```diff
@@ -15,7 +15,7 @@ fun Event.toAlarmItem(agendaType: AgendaType): AlarmItem {
         agendaId = this.id,
         title = this.title,
         description = this.description,
-        remindAt = this.remindAt * 1000,
+        remindAtInMilliSecs = this.remindAt * 1000,
```
steve1rm: @philipplackner Is it better to change the name here to something more readable.

 As the `setExact` needs it in milli seconds
```
alarmManager.setExact(
                AlarmManager.RTC_WAKEUP,
                alarmItem.remindAtInMilliSecs
               ......)
```

Or leave the name as it is `remindAt` and just do the following here
```
alarmManager.setExact(
                AlarmManager.RTC_WAKEUP,
                alarmItem.remindAt * 1000
               ......)
```

MKiperszmid: What I prefer is to have remindAt be the zoneddatetime, and then on the alarm manager convert it to milliseconds

Mostly so you always know you'll receive a valid date on the AlarmItem

presentation/src/main/java/me/androidbox/presentation/BusbyTaskyApplication.kt
```diff
@@ -1,7 +1,45 @@
...
+        createNotificationChannel(generateChannels())
+    }
+
+    private fun createNotificationChannel(listOfChannel: Map<String, String>) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            listOfChannel.map { mapOfChannel ->
+                val notificationManager = getNotificationManager()
+                val notificationChannel = NotificationChannel(
+                    mapOfChannel.key,
+                    mapOfChannel.value,
```
MKiperszmid, line 25: This is correct

But
When you check the key and values on the map, they are the other way wrong

Key should be "event", "reminder", "task", and value should have the description
But on the map creation, you are doing it the opposite way

presentation/src/main/java/me/androidbox/presentation/BusbyTaskyApplication.kt
MKiperszmid: Here, read comment above

MKiperszmid (approved):

domain/src/main/java/me/androidbox/domain/alarm_manager/AlarmItem.kt
steve1rm: @philipplackner what do you think. I do perfer @MKiperszmid as in my second suggestion. And convert it to millis seconds only when needed.

presentation/src/main/java/me/androidbox/presentation/BusbyTaskyApplication.kt
steve1rm: @MKiperszmid Thanks for catching that. I have updated

steve1rm:
@MKiperszmid @philipplackner Decided to go with @MKiperszmid suggestion about converting to millis seconds only when its needed.

PR #42, opened 2023-04-10
[40] Verify that a visitor exists

steve1rm:
# Pull Request template

###  What does this PR solve?

- Show visitor dialog box
- Adds endpoint for verifying if the attendee exists on the BE
- Checks whether the response contains a valid visitor or not.

addvisitor


presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -158,6 +163,31 @@ fun EventScreen(
...
+                    eventScreenEvent(EventScreenEvent.OnShowVisitorDialog(shouldShowVisitorDialog = false))
+                },
+                isValidInput = false,
+                onAddButtonClicked = { visitorEmail ->
+                    eventScreenEvent(EventScreenEvent.CheckVisitorExists(visitorEmail))
+                },
+                isEmailVerifiedSuccess = eventScreenState.isEmailVerifiedSuccess
+            )
+        }
+    }
```
steve1rm, line 188: Not sure why this is a problem. But when there is an error I display the "Email does not exist". I don't think there is any issue with the AddVisitorDialog as the previews seem to show ok. Could it be something to do with the `Box` As the button is half cut-off

addvisitor-error

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -133,6 +140,72 @@ class EventViewModel @Inject constructor(
...
+                is ResponseState.Success -> {
+                    responseState.data?.let { _attendee ->
+                        val attendee = _attendee.copy(
+                            email = _attendee.email,
+                            fullName = _attendee.fullName,
+                            userId = _attendee.userId,
+                            remindAt = _attendee.remindAt,
+                            eventId = eventScreenState.value.eventId,
+                            isGoing = _attendee.isGoing
+                        )
```
steve1rm, line 182: I am just wondering what should `remindAt` should be as we don't get that from the endpoint when we verify the visitors email:

```
{
"doesUserExist": true/false, (Boolean)
"attendee": {
"email": "<Attendee email>", (String)
"fullName": "<Attendee full name>", (String)
"userId": "<Attendee user ID>" (String)
}
}
```

steve1rm: Another question I have is if the user clicks on `Add` button multiple times. What is the best way to handle this as it would make multiple requests?

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -133,6 +140,72 @@ class EventViewModel @Inject constructor(
...
+
+                        _eventScreenState.update { eventScreenState ->
+                            eventScreenState.copy(
+                                isEmailVerifiedSuccess = true,
+                                listOfAttendee = eventScreenState.listOfAttendee + attendee
+                            )
+                        }
+
+                        /* TODO Add this attendee to the event and update this event in db */
```
steve1rm: When we get a valid response back with an visitor.
1) This attendee will be added to the Event, and that Event will be updated in the room DB with the added attendee?
2) We have call our work manager to upload that event as well. As we did for the insertion of a new event?

component/src/main/java/me/androidbox/component/agenda/VisitorFilter.kt
```diff
@@ -21,14 +25,30 @@ import java.util.*
 fun VisitorFilter(
     modifier: Modifier = Modifier,
     selectedVisitorType: VisitorType,
```
MKiperszmid, line 26: Remember modifier should be the first optional paraemeter
selectedVisitorType isn't optional, so it should be before the modifier

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -158,6 +163,31 @@ fun EventScreen(
         }
     )
 
+    if(eventScreenState.shouldShowVisitorDialog) {
+        Box(modifier = Modifier.fillMaxSize()) {
```
MKiperszmid: Do you really need this box tho?

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
MKiperszmid: Maybe using AlertDialog instead of Dialog?

Also, Box isn't really required, since Dialog already has a Box underneath (if I'm not mistaken)

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
```diff
@@ -8,13 +8,18 @@ import me.androidbox.component.event.VisitorInfo
...
 import java.time.ZonedDateTime
+import java.util.UUID
 
 data class EventScreenState(
     val listOfPhotoUri: List<String> = mutableStateListOf<String>(),
     val selectedVisitorType: VisitorType = VisitorType.ALL,
     val selectedVisitor: VisitorInfo? = null,
+    val visitorEmail: String = "",
+    val shouldShowVisitorDialog: Boolean = false,
+    val isEmailVerifiedSuccess: Boolean = true,
```
MKiperszmid: Do you need the word "Success" here?
isEmailVerified should be enough :D

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -75,6 +77,11 @@ class EventViewModel @Inject constructor(
                 }
             }
             is EventScreenEvent.OnSaveEventDetails -> {
+                _eventScreenState.update { eventScreenState ->
+                    eventScreenState.copy(
+                        eventId = UUID.randomUUID().toString()
+                    )
```
MKiperszmid, line 83: On your state, you are already setting the eventId

What's the difference between both?

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
MKiperszmid: Regarding the first question:
You don't care about remindAt of other people. In fact, you won't use it anywhere in the app.

Second question: You could have a "isLoading", and whenever it's false, then you make the API call and set it to true, and once it's done you set it back to false

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
MKiperszmid: For both questions:
You will do it when you save the event, not when you validate a visitor

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
steve1rm: @MKiperszmid So this is enough
`val eventId: String = UUID.randomUUID().toString(),`
And I can remove this: 
```
  eventScreenState.copy(
      eventId = UUID.randomUUID().toString()
 )
```

data/src/main/java/me/androidbox/data/remote/event/VerifyVisitorEmailUseCaseImp.kt
```diff
@@ -0,0 +1,36 @@
...
+import me.androidbox.data.mapper.toAttendee
+import me.androidbox.data.remote.model.response.AttendeeEmailVerifyResponseDto
+import me.androidbox.data.remote.network.event.EventService
+import me.androidbox.data.remote.util.CheckResult.checkResult
+import me.androidbox.domain.authentication.ResponseState
+import me.androidbox.domain.authentication.model.Attendee
+import me.androidbox.domain.event.usecase.VerifyVisitorEmailUseCase
+import javax.inject.Inject
+
+class VerifyVisitorEmailUseCaseImp @Inject constructor(
```
philipplackner, line 12: The data layer is neither the place for use cases nor are use cases the place to put data-related calls. Use cases should be completely isolated from 3rd party SDKs or the Android SDK and just contain business logic or work with abstractions from the domain layer (such as your repo interface).
Also, there is no point in abstracting a use case since you'll never have a second implementation of it.

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
philipplackner: Would also be my guess that `AlertDialog` has a fixed height while `Dialog` does not

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
philipplackner: Yes, after an attendee is added via the dialog, you just add it to your local event state of your ViewModel. Once you hit save, you upload the event with all the added attendees. 
I'd recommend to use WorkManager to upload an event as well, since it might contain up to 10 photos which could take a bit

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
steve1rm: @philipplackner I am using the Dialog. However, I have removed the Box, but still has the same behavior.

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
steve1rm: @MKiperszmid Updated

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
steve1rm: @MKiperszmid Removed the box but the behaviour is the same

PR #43, opened 2023-04-12
[41] Open Event Detail screen, delete event

steve1rm:
# Pull Request template

###  What does this PR solve?
Shows dropdown menu of actions to have related event (open, edit, delete)
Can delete agenda items
Can open the event detail screen

ERROR:
When fetching the event from the DB in the navigation graph it will fetch the event from the DB again and overwrite any state that was updated when coming back from the Edit screen.

https://user-images.githubusercontent.com/14260802/231448808-756155b3-94ca-4597-9a7e-09cd3f52cd63.mp4

### DIFF
--- a/presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
+++ b/presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
@@ -32,7 +32,7 @@ import me.androidbox.component.ui.theme.backgroundBackColor
 import me.androidbox.component.ui.theme.dropDownMenuBackgroundColor
 import me.androidbox.domain.DateTimeFormatterProvider.toDisplayDateTime
 import me.androidbox.domain.DateTimeFormatterProvider.toZoneDateTime
-import me.androidbox.domain.agenda.model.AgendaMenuActionType
+import me.androidbox.presentation.agenda.constant.AgendaMenuActionType
 import me.androidbox.domain.alarm_manager.AgendaType
 import me.androidbox.presentation.ui.theme.BusbyTaskyTheme
 
@@ -139,7 +139,7 @@ fun AgendaScreen(
                             agendaScreenEvent(
                                 AgendaScreenEvent.OnChangeShowEditAgendaItemDropdownStatus(shouldOpen = false))
                         },
-                        listOfMenuItemId = listOf(R.string.open, R.string.edit, R.string.delete),
+                        listOfMenuItemId = AgendaMenuActionType.values().map { it.titleId },
                         onSelectedOption = { item ->
                             agendaScreenState.agendaItemClicked?.let { agendaItem ->
                                 onSelectedEditAgendaItemClicked(agendaItem.id, agendaItem.agendaType, AgendaMenuActionType.values()[item])
--- a/presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
+++ b/presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
@@ -40,6 +40,7 @@ import me.androidbox.presentation.ui.theme.BusbyTaskyTheme
 fun AgendaScreen(
     agendaScreenState: AgendaScreenState,
     agendaScreenEvent: (AgendaScreenEvent) -> Unit,
+    onSelectedEditAgendaItemClicked: (id: String, AgendaType) -> Unit,
     onSelectedAgendaItem: (agendaType: Int) -> Unit, /* TODO Check where this is being used */
     modifier: Modifier = Modifier) {
 
@@ -124,7 +125,27 @@ fun AgendaScreen(
                         isAgendaCompleted = false
                     ) {
                         println("Event ${agendaItem.id} has been clicked")
+                        agendaScreenEvent(AgendaScreenEvent.OnAgendaItemClicked(agendaItem))
+                        agendaScreenEvent(AgendaScreenEvent.OnChangeShowEditAgendaItemDropdownStatus(shouldOpen = true))
                     }
+
+                    AgendaDropDownMenu(
+                        modifier = Modifier
+                            .background(color = MaterialTheme.colorScheme.dropDownMenuBackgroundColor)
+                            .align(Alignment.BottomEnd),
+                        shouldOpenDropdown = agendaScreenState.shouldOpenEditAgendaDropdown,
+                        onCloseDropdown = {
+                            agendaScreenEvent(
+                                AgendaScreenEvent.OnChangeShowEditAgendaItemDropdownStatus(shouldOpen = false))
+                        },
+                        listOfMenuItemId = listOf(R.string.open, R.string.edit, R.string.delete),
+                        onSelectedOption = { item ->
+                            agendaScreenState.agendaItemClicked?.let { agendaItem ->
+                                onSelectedEditAgendaItemClicked(agendaItem.id, agendaItem.agendaType)
+                                agendaScreenEvent(AgendaScreenEvent.OnChangeShowEditAgendaItemDropdownStatus(shouldOpen = false))
+                            }
+                        }
+                    )
                 }
             }
         }
@@ -154,6 +175,7 @@ fun PreviewAgendaScreen() {
                 .background(
                     color = MaterialTheme.colorScheme.agendaBackgroundColor),
             agendaScreenEvent = {},
+            onSelectedEditAgendaItemClicked = { _, _ -> },
             onSelectedAgendaItem = {}
         )
     }
--- a/presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
+++ b/presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
@@ -32,6 +32,7 @@ import me.androidbox.component.ui.theme.backgroundBackColor
 import me.androidbox.component.ui.theme.dropDownMenuBackgroundColor
 import me.androidbox.domain.DateTimeFormatterProvider.toDisplayDateTime
 import me.androidbox.domain.DateTimeFormatterProvider.toZoneDateTime
+import me.androidbox.domain.agenda.model.AgendaMenuActionType
 import me.androidbox.domain.alarm_manager.AgendaType
 import me.androidbox.presentation.ui.theme.BusbyTaskyTheme
 
@@ -40,7 +41,7 @@ import me.androidbox.presentation.ui.theme.BusbyTaskyTheme
 fun AgendaScreen(
     agendaScreenState: AgendaScreenState,
     agendaScreenEvent: (AgendaScreenEvent) -> Unit,
-    onSelectedEditAgendaItemClicked: (id: String, AgendaType) -> Unit,
+    onSelectedEditAgendaItemClicked: (id: String, AgendaType, agendaMenuActionType: AgendaMenuActionType) -> Unit,
     onSelectedAgendaItem: (agendaType: Int) -> Unit, /* TODO Check where this is being used */
     modifier: Modifier = Modifier) {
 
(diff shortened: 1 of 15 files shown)


presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -206,6 +207,32 @@ class EventViewModel @Inject constructor(
...
+                                eventDescription = event.description,
+                            )
+                        }
+                    }
+                    is ResponseState.Failure -> {
+                        Log.d("EVENT_DETAIL", "${responseState.error.message}")
+                    }
+                }
+            }
+        }
```
steve1rm, line 273: Get the event from the DB and updates the state. I am just testing with the title and description at the moment.

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
```diff
@@ -131,6 +153,15 @@ fun NavigationGraph(
...
             }
 
+            /* FIXME: Issue when landing back from the Edit Screen this will run
+            *         and overwrite the state that was input in the Edit Screen
+            *         and it will fetch from the DB and restore the state from there */
+            it.arguments?.getString(Screen.EventScreen.EVENT_ID)?.let { eventId ->
+                LaunchedEffect(key1 = true) {
+                    eventViewModel.fetchEventById(eventId)
+                }
+            }
```
steve1rm: @philipplackner @MKiperszmid I think this is the issue. As will run to get the event from the DB. So this event screen will run when navigating back from the Edit Screen and also when opening from the Agenda Screen. 

What is the best way to handle this case?

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
```diff
@@ -124,7 +126,27 @@ fun AgendaScreen(
...
+                    AgendaDropDownMenu(
+                        modifier = Modifier
+                            .background(color = MaterialTheme.colorScheme.dropDownMenuBackgroundColor)
+                            .align(Alignment.BottomEnd),
+                        shouldOpenDropdown = agendaScreenState.shouldOpenEditAgendaDropdown,
+                        onCloseDropdown = {
+                            agendaScreenEvent(
+                                AgendaScreenEvent.OnChangeShowEditAgendaItemDropdownStatus(shouldOpen = false))
+                        },
+                        listOfMenuItemId = listOf(R.string.open, R.string.edit, R.string.delete),
```
MKiperszmid: Didn't you create an AgendaMenuActionType for this?
You can use those, and have the Enum receive the ids as params

So smth like:

```
enum class AgendaMenuActionType(@StringRes val id: Int) {
    OPEN(R.string.open),
    EDIT(R.string.edit),
    DELETE(R.string.delete)
}

And then use it like:

AgendaMenuActionType.values() to get a list of AgendaMenuActionType

File: presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt (hunk 2 above)
Comment (MKiperszmid) [reply presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt]: Why are you using a LaunchedEffect to get the event by id? If the screen gets recomposed, it will get the event again

You can just call it on the viewmodel's init block
That won't get recomposed, so it should work

**Review by MKiperszmid — APPROVED** [review]

File: presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt (hunk 3 above)
Comment (steve1rm) [reply presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt]: @MKiperszmid I will update

File: presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt (hunk 2 above)
Comment (steve1rm) [reply presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt]: This is what I am doing. The user taps on one of the agenda cards i.e. events
1) Dropdown menu will open and the user selected `Open`
2) I send this event to update the state which contains the agenda item has has been clicked: 
`agendaScreenEvent(AgendaScreenEvent.OnAgendaItemClicked(agendaItem))`
3) In my AgendaViewModel I update the state for the agenda item that has been selected
```
            is AgendaScreenEvent.OnAgendaItemClicked -> {
                _agendaScreenState.update { agendaScreenState ->
                    agendaScreenState.copy(
                        agendaItemClicked = agendaScreenEvent.agendaItem
                    )
                }
            }
```
4) In my AgendaScreen I have this lamba 
`onSelectedEditAgendaItemClicked: (id: String, AgendaType, agendaMenuActionType: AgendaMenuActionType) -> Unit`

4) The result of this is used in the NavigationGraph to send the event ID argument to the EventScreen
```
 AgendaMenuActionType.OPEN -> {
                                navHostController.navigate(route = "${Screen.EventScreen.EVENT_SCREEN}/$id")
                            }
```
5) I use the LaunchEffect to fetch the event ID from the arguments so this event can be fetched from the DB
```
 it.arguments?.getString(Screen.EventScreen.EVENT_ID)?.let { eventId ->
                LaunchedEffect(key1 = true) {
                    eventViewModel.fetchEventById(eventId)
                }
            }
```

Basically, I need to send the event ID from the Agenda Screen to open the EventDetailScreen by fetching the event from the DB. The issue having when navigating back from the EditDetailScreen and fetching the same event ID from the DB.

Comment (MKiperszmid) [reply presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt]: I'm not saying that logic is bad

I'm not really understanding why you are calling fetchEventById, on the composable, and not directly on the EventViewModel's init block

Comment (steve1rm) [reply presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt]: @MKiperszmid I don't think moving the logic will work if I move it to the init {}. The ViewModel will only get initialized when created just one time. So that init {} will only run once. 
When the user clicks on the agenda card to open it. I use the eventID to fetch from the DB. But as the composable will re-compose that is called again in the navigationGraph. 
I have updated my code with another solution. But that need not work either.

Comment (MKiperszmid) [reply presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt]: You should pass the id over navArguments, and grab it using savedStateHandle
That's where you can then call the getEventByid using the id from there

Comment (MKiperszmid) [reply presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt]: This is how I did it:



```
 composable(
            route = Route.EDITOR + "/{text_id}/{title}/{body}/{size}",
            arguments = listOf(
                navArgument("body") {
                    type = NavType.StringType
                },
                navArgument("title") {
                    type = NavType.StringType
                },
                navArgument("size") {
                    type = NavType.IntType
                },
                navArgument("text_id") {
                    type = NavType.StringType
                }
            )
        ) {
            val id = it.arguments?.getString("text_id") ?: ""
            EditorScreen(
                title = it.arguments?.getString("title") ?: "Edit Text",
                onBack = {
                    navController.previousBackStackEntry?.savedStateHandle?.set(id, null)
                    navController.popBackStack()
                },
                onSave = { body ->
                    navController.previousBackStackEntry?.savedStateHandle?.set(id, body)
                    navController.popBackStack()
                },
                textSize = (it.arguments?.getInt("size") ?: 26).sp
            )
        }

composable(
            route = Route.TASK + "?date={date}&action={action}&id={id}",
            arguments = listOf(
                navArgument("date") {
                    type = NavType.StringType
                    nullable = true
                    defaultValue = null
                },
                navArgument("action") {
                    type = NavType.StringType
                    nullable = true
                    defaultValue = null
                },
                navArgument("id") {
                    type = NavType.StringType
                    nullable = true
                    defaultValue = null
                }
            ),
            deepLinks = listOf(navDeepLink { uriPattern = DeepLinks.TASK_DETAIL })
        ) {
            val taskTitle = it.savedStateHandle.get<String>("task_title") ?: ""
            val taskDescription = it.savedStateHandle.get<String>("task_description") ?: ""
            DetailTaskScreen(
                taskTitle = taskTitle,
                taskDescription = taskDescription,
                onClose = {
                    navController.navigateUp()
                },
                openEditor = { id, title, body, size ->
                    navController.navigate(Route.EDITOR + "/$id/$title/$body/$size")
                }
            )
        }
```

Comment (MKiperszmid) [reply presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt]: Then on the Task screen I have:
```
 LaunchedEffect(taskTitle, taskDescription) {
        viewModel.onEvent(
            DetailTaskEvent.OnUpdatedInformation(
                taskTitle,
                taskDescription
            )
        )
    }
```

Comment (MKiperszmid) [reply presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt]: And the viewmodel makes the getTaskById when you receive the id over navArguments:
```
  init {
val itemId = savedStateHandle.get<String>("id")
        if (itemId != null) {
...
```

File: presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt [hunk 4]
```diff
@@ -39,6 +39,10 @@ class EventViewModel @Inject constructor(
     private val _eventScreenState: MutableStateFlow<EventScreenState> = MutableStateFlow(EventScreenState())
     val eventScreenState = _eventScreenState.asStateFlow()
 
+    init {
+        Log.d("EVENT_VIEWMODEL", "THIS IS IN THE EVENT_VIEW_MODEL")
+        eventScreenState.value.eventId
```
MKiperszmid: Nono
You should receive the savedStateHandle in the viewmodel constructor
And then, get the id as I showed you on my code

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
```diff
@@ -157,9 +157,7 @@ fun NavigationGraph(
             *         and overwrite the state that was input in the Edit Screen
             *         and it will fetch from the DB and restore the state from there */
             it.arguments?.getString(Screen.EventScreen.EVENT_ID)?.let { eventId ->
-                LaunchedEffect(key1 = true) {
-                    eventViewModel.fetchEventById(eventId)
-                }
+                eventViewModel.onEventScreenEvent(EventScreenEvent.OnSaveEventID(eventId))
```
MKiperszmid: Remove this line. It will be done in the viewmodel

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -32,12 +34,19 @@ class EventViewModel @Inject constructor(
...
+    private val uploadEvent: UploadEvent,
+    savedStateHandle: SavedStateHandle
 ) : ViewModel() {
 
     private val _eventScreenState: MutableStateFlow<EventScreenState> = MutableStateFlow(EventScreenState())
     val eventScreenState = _eventScreenState.asStateFlow()
 
+    init {
+        savedStateHandle.get<String>("event_id")?.let { eventId ->
+            eventRepository.getEventById(eventId)
```
MKiperszmid: You should set the event's screen state information with this event you get from the repository

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -159,12 +168,14 @@ class EventViewModel @Inject constructor(
             is EventScreenEvent.CheckVisitorExists -> {
                 verifyVisitorEmail(eventScreenEvent.visitorEmail)
             }
+            is EventScreenEvent.OnSaveEventID -> {
+                fetchEventById(eventScreenEvent.eventId)
+            }
```
MKiperszmid: Why do you need this?
You named it "OnSaveEvent" meaning it should save, but you are fetching it

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
steve1rm: @MKiperszmid This has been removed already. It was something I was trying when resolving this issue

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
steve1rm: @MKiperszmid Updated this already. I am not calling a function that collect this state from the DB

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
steve1rm: @MKiperszmid Been removed already

PR #44, opened 2023-04-12
[42] get agenda specific day and inserting 

steve1rm:
# Pull Request template

###  What does this PR solve?

- Creates the endpoint for the `/agenda` endpoint
- Fetches the agenda for the specific day
- Inserts the events, tasks, and reminders into the DB
- Fetches all the events, tasks, and reminders from the DB


data/src/main/java/me/androidbox/data/remote/agenda/AgendaRemoteRepositoryImp.kt
```diff
@@ -0,0 +1,41 @@
...
+            reminderDto.toReminder()
+        }
+
+        return Agenda(
+            events = events,
+            tasks = tasks,
+            reminders = reminders
+        )
+    }
+}
```
steve1rm, line 41: Fetches all the agenda items form the `/agenda` endpoint. Maps the Dto to the models

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
```diff
@@ -0,0 +1,94 @@
...
+                tasks = tasks.single(),
+                reminders = reminders.single()
+            )
+
+            emit(ResponseState.Success(fullAgenda))
+        }
+            .catch { throwable ->
+                emit(ResponseState.Failure(throwable))
+            }
+    }
```
steve1rm, line 127: @philipplackner @MKiperszmid I am not sure if this is the best way to handle flows like this. If this is incorrect, please give a code snippet :blush:

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
```diff
@@ -0,0 +1,94 @@
...
+            )
+
+            /* Insert events tasks and reminders into the DB */
+            supervisorScope {
+                agenda.events.map { event ->
+                    launch {
+                        eventDao.insertEvent(event.toEventEntity())
+                    }
+                }.forEach { job -> job.join() }
+            }
```
MKiperszmid: Another solution here could be to create a "insertEvents" on the dao, that accepts a list of event entity

Same for task and repository

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
```diff
@@ -0,0 +1,94 @@
...
+
+class AgendaLocalRepositoryImp @Inject constructor(
+    private val eventDao: EventDao,
+    private val taskDao: TaskDao,
+    private val reminderDao: ReminderDao,
+    private val agendaRemoteRepository: AgendaRemoteRepository
+) : AgendaLocalRepository {
+
+    override fun fetchAgenda(startTimeStamp: Long, endTimeStamp: Long): Flow<ResponseState<Agenda>> {
+        return flow<ResponseState<Agenda>> {
```
MKiperszmid, line 31: You are using Flow, yet only emiting the value just once (at the end). So you aren´t really taking advantage of Flows over a regular method

One way to take advantage would be to do:

Get items from the DB and emit them
Get items from API
Save to db
Emit those items

Most of the times, you´ll emit the same values (which if you use flow´s "Distinct", nothing´s gonna happen), but if you get stuff from API that are new, they will be saved on the db and then emitted

This is important, since right now you are calling the API, and if the API takes 5 seconds for some reason, you aren´t displaying any items to the user

The whole point of an offline first is to show the user info, even if the connectivity is slow or non existant

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
steve1rm: @MKiperszmid @philipplackner I understand the concept that we first want to show the user the agenda items before fetching from the API. And if there is items we should insert them, and emit them again. However, I am having issues with the coding part of this using flows. That is were I need help with.

I think I am doing something wrong with my AgendaLocalRepositoryImp class when I am fetching and emitting. Please help

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
```diff
@@ -0,0 +1,124 @@
...
+                events = events,
+                tasks = tasks,
+                reminders = reminders
+            )))
+
+            /* Fetch all agenda for the current day and time zone */
+            val agenda = agendaRemoteRepository.fetchAgendaForDay(
+                ZoneId.systemDefault(),
+                startTimeStamp * 1000
+            )
```
steve1rm, line 49: @MKiperszmid I think there is something wrong. As I have events for the date I am fetching for. But the endpoint always returns empty. Is this correct?

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
MKiperszmid: I see you added the code that does this. Still need help?

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
MKiperszmid: Do you have stuff on the api?
If so, is the timeStamp already in ms? if so, try removing the "* 1000"

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
steve1rm: @MKiperszmid Yes, changing to suspend functions resolved that issue. But not sure why the flow didn't work. Maybe that is a questions to @philipplackner

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
steve1rm: I use this method to get the start of the current day?   
```
private fun getStartOffCurrentDay(agendaDate: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault())): ZonedDateTime {
        return agendaDate
            .toLocalDate()
            .atStartOfDay(ZoneId.systemDefault())
    }
```

Then I call it like this:
`agendaLocalRepository.fetchAgenda(getStartOffCurrentDay(agendaDate).toEpochSecond(), getEndOfCurrentDay(agendaDate).toEpochSecond())`

Which I use toEpochSecond() to change it to seconds

Then I change it to milli seconds by multipying by 1000 
```
            val agenda = agendaRemoteRepository.fetchAgendaForDay(
                ZoneId.systemDefault(),
                startTimeStamp * 1000
            )

I should have items on the BE for that day as they have already been upload every time I save to the db I use work manager to upload them. 
```

MKiperszmid: Have you verified this, by checking the API with something like Postman?
Mostly to verify that you are indeed not getting any items from the API because an issue with your code, and not that the api is empty

steve1rm: Screenshot from 2023-04-14 02-18-18

steve1rm: Screenshot from 2023-04-14 02-18-18

MKiperszmid: Then issue is that you aren't uploading the information correctly.
Have you tried uploading a Task or a Reminder just to verify it works?

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
philipplackner: @steve1rm Maybe you didn't launch the flow in the ViewModel by using `collectLatest` or `launchIn`?

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
```diff
@@ -0,0 +1,124 @@
...
+                }.forEach { job -> job.join() }
+            }
+
+            supervisorScope {
+                agenda.reminders.map { reminder ->
+                    launch {
+                        reminderDao.insertReminder(reminder.toReminderEntity())
+                    }
+                }.forEach { job -> job.join() }
+            }
```
philipplackner: With this approach you insert all events in parallel, then insert all tasks in parallel and then all reminders. You could do it all in parallel though by first just mapping to the jobs for all types of items and then joining all

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    private suspend fun fetchTasksFromLocal(
+        startTimeStamp: Long,
+        endTimeStamp: Long
+    ): List<Task> {
+        val tasks = taskDao.getTasksFromTimeStampFullAgenda(startTimeStamp, endTimeStamp)
+            .map { taskEntity ->
+                taskEntity.toTask()
+            }
+        return tasks
```
philipplackner, line 115: If you just create a variable to return it afterwards, I'd just leave it

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
philipplackner: Looks good

data/src/main/java/me/androidbox/data/remote/agenda/AgendaRemoteRepositoryImp.kt
philipplackner: Could it make sense to return the agenda here as a `List<AgendaItem>` sorted by time?

philipplackner (approved):

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
steve1rm, line 1: Update to run all insertions in parallet

philipplackner (approved):

PR #45, opened 2023-04-16
[44] Adds entity for events, tasks, and reminders for sync'ing

steve1rm:
# Pull Request template

###  What does this PR solve?

- Adds the insertion for sync'ing the event agenda item for CREATE and DELETE
- Entities for Task, Event, and Reminder

### DIFF
--- a/data/src/main/java/me/androidbox/data/local/event/EventRepositoryImp.kt
+++ b/data/src/main/java/me/androidbox/data/local/event/EventRepositoryImp.kt
@@ -53,6 +53,7 @@ class EventRepositoryImp @Inject constructor(
 
         val result = checkResult {
             eventDao.insertEvent(eventEntity)
+            eventDao.insertSyncEvent(EventSyncEntity(event.id, SyncAgendaType.CREATE))
         }
 
         val responseState = result.fold(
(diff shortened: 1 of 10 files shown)


data/src/main/java/me/androidbox/data/local/event/EventRepositoryImp.kt
```diff
@@ -68,6 +71,7 @@ class EventRepositoryImp @Inject constructor(
     override suspend fun deleteEventById(id: String): ResponseState<Unit> {
         val result = checkResult {
             eventDao.deleteEventById(id)
+            eventDao.insertSyncEvent(EventSyncEntity(id, SyncAgendaType.DELETE))
```
steve1rm, line 74: @MKiperszmid @philipplackner Is it ok to run these one after another.

data/src/main/java/me/androidbox/data/local/event/EventRepositoryImp.kt
```diff
@@ -51,6 +53,7 @@ class EventRepositoryImp @Inject constructor(
 
         val result = checkResult {
             eventDao.insertEvent(eventEntity)
+            eventDao.insertSyncEvent(EventSyncEntity(event.id, SyncAgendaType.CREATE))
```
steve1rm, line 56: @MKiperszmid @philipplackner Is it ok to run these one after another.

philipplackner: You could use a transaction to only execute both if both succeed. You can do this by using `db.withTransaction`. Very unlikely inserting into a DB fails, but theoretically possible

philipplackner (approved):

PR #46, opened 2023-04-16
[43] sync fullagenda syncaganda

steve1rm:
# Pull Request template

###  What does this PR solve?

- Runs the SyncAgenda work manager.
- Makes successful request


data/src/main/java/me/androidbox/data/remote/network/agenda/AgendaService.kt
```diff
@@ -11,4 +15,10 @@ interface AgendaService {
     suspend fun getAgendaForDay(
         @Query("timezone") zoneId: ZoneId,
         @Query("time") time: Long): AgendaDto
+
+    @POST("/syncAgenda")
+    suspend fun syncAgenda(@Body syncAgenda: SyncAgendaDto)
```
steve1rm, line 20: For this request do we need to return anything in the response?

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsImp.kt
```diff
@@ -0,0 +1,60 @@
...
+import me.androidbox.domain.work_manager.SyncAgendaItems
+import java.util.UUID
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+class SyncAgendaItemsImp @Inject constructor(
+    private val workManager: WorkManager,
+    private val eventDao: EventDao,
+    private val taskDao: TaskDao,
+    private val reminderDao: ReminderDao
```
steve1rm: I am passing these as parameters and then fetching the ids and then passing them as `workDateOf` to the workmanager. But would it be better to just pass these  `eventDao`s directly in the constructor of the `SyncAgendaItemsWorker()` Which is better?

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -130,4 +144,29 @@ class AgendaViewModel @Inject constructor(
...
+                /* Delete the ids from the deleted sync table */
+                viewModelScope.launch {
+                    agendaLocalRepository.deleteEventSyncType(SyncAgendaType.DELETE)
+                }
+            }
+        }
+
+        val syncAgendaId = viewModelScope.async {
+            syncAgendaItems.sync()
+        }
```
steve1rm, line 144: I call this in `async` block as I want to wait for the result of the UUID from the worker

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -130,4 +144,29 @@ class AgendaViewModel @Inject constructor(
...
+
+        viewModelScope.launch {
+            workManagerLiveData = workManager.getWorkInfoByIdLiveData(syncAgendaId.await())
```
steve1rm: I use `await` so I have to use launch here as `await` is a suspending function.

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -130,4 +144,29 @@ class AgendaViewModel @Inject constructor(
             }
         }
     }
+
+    private fun syncAgendaItems() {
+        agendaSyncObserver = Observer { workInfo ->
+            if(workInfo.state == WorkInfo.State.SUCCEEDED) {
+                /* Delete the ids from the deleted sync table */
+                viewModelScope.launch {
+                    agendaLocalRepository.deleteEventSyncType(SyncAgendaType.DELETE)
```
steve1rm: As `agendaLocalRepository.deleteEventSyncType(SyncAgendaType.DELETE)` doesn't return any result I can just use launch block

steve1rm: An issue with this, is that this code never seems to run.

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -130,4 +144,29 @@ class AgendaViewModel @Inject constructor(
...
+
+        val syncAgendaId = viewModelScope.async {
+            syncAgendaItems.sync()
+        }
+
+        viewModelScope.launch {
+            workManagerLiveData = workManager.getWorkInfoByIdLiveData(syncAgendaId.await())
+            workManagerLiveData.observeForever(agendaSyncObserver)
+        }
+    }
```
steve1rm: I think there could be many issues with this `syncAgendaItems`  Please provide some code snippets to help provide some clarity.

data/src/main/java/me/androidbox/data/remote/network/agenda/AgendaService.kt
MKiperszmid: A success or failure
If success, remove the deletedEventIds, deletedTaskId and deletedReminder ids from the database

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsImp.kt
MKiperszmid: You could just inject the repository on the workmanager directly, and call a "getAllDeletedItems" or similar
No need to pass them using workerInfo or anything

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -130,4 +144,29 @@ class AgendaViewModel @Inject constructor(
             }
         }
     }
+
+    private fun syncAgendaItems() {
+        agendaSyncObserver = Observer { workInfo ->
+            if(workInfo.state == WorkInfo.State.SUCCEEDED) {
```
MKiperszmid: If you pass the repository directly to the worker, all of this can be removed

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsImp.kt
```diff
@@ -0,0 +1,42 @@
...
+            .setConstraints(Constraints(Constraints.Builder().setRequiredNetworkType(NetworkType.CONNECTED).build()))
+            .setBackoffCriteria(
+                BackoffPolicy.EXPONENTIAL,
+                WorkRequest.MIN_BACKOFF_MILLIS,
+                TimeUnit.MILLISECONDS)
+            .build()
+
+        workManager.enqueueUniquePeriodicWork(
+            SYNC_DELETED_AGENDA_IDS,
+            ExistingPeriodicWorkPolicy.KEEP,
```
steve1rm: Is this the correct policy or should I use `CANCEL_AND_REENQUEUE`?

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsWorker.kt
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            deletedTaskIds = async {
+                taskDao.getAllTasksBySyncType(SyncAgendaType.DELETE)
+            }
+
+            deletedReminderIds = async {
+                reminderDao.getAllRemindersBySyncType(SyncAgendaType.DELETE)
+            }
+        }
```
steve1rm, line 51: Is this correct using supervisorScope to get all the ids from the DB?

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsWorker.kt
```diff
@@ -0,0 +1,84 @@
...
+        return result
+    }
+
+    private suspend fun deleteAllDeletedSyncAgendaItems() {
+        withContext(Dispatchers.IO) {
+            eventDao.deleteSyncEventsBySyncType(SyncAgendaType.DELETE)
+            taskDao.deleteSyncTasksBySyncType(SyncAgendaType.DELETE)
+            reminderDao.deleteSyncRemindersBySyncType(SyncAgendaType.DELETE)
+        }
+    }
```
steve1rm, line 95: Is this correct using withContext() or should this be supervisiorScope as well?

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
steve1rm: @MKiperszmid and @philipplackner Just for the sack of learning. Is my `syncAgendaItems` correct. If not how can I make this correct? 

I have already made the changes as requested by @MKiperszmid to pass in the dao to the worker. But just interested to know if I am doing something wrong with the above code.

data/src/main/java/me/androidbox/data/remote/network/agenda/AgendaService.kt
philipplackner: Therefore, no you don't need to return anything. If it fails, it'll throw an exception which you can catch

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsImp.kt
philipplackner: I think it should just be replace, right? Not sure what the difference between replace and cancel and reenqueue is, but both sound better than Keep. If you use keep, you'll have multiple workers running trying to sync the same agenda items

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsWorker.kt
philipplackner: Yes that works

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsWorker.kt
```diff
@@ -0,0 +1,84 @@
...
+            deletedReminderIds = async {
+                reminderDao.getAllRemindersBySyncType(SyncAgendaType.DELETE)
+            }
+        }
+
+        val responseResult = checkResult {
+            val syncAgendaDto = SyncAgendaDto(
+                deletedEventIds = deletedEventIds.await().toList(),
+                deletedTaskIds = deletedTaskIds.await().toList(),
+                deletedReminderIds = deletedReminderIds.await().toList())
```
philipplackner: These `toList()` shouldn't be necessary here

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsWorker.kt
philipplackner: withContext works, but you could also run them all in parallel in separate `launch` blocks

domain/src/main/java/me/androidbox/domain/work_manager/SyncAgendaItems.kt
```diff
@@ -0,0 +1,7 @@
+package me.androidbox.domain.work_manager
+
+import java.util.UUID
+
+interface SyncAgendaItems {
+    suspend fun sync(): UUID
+}
```
philipplackner: I'd avoid calling interfaces like functions, since they're not. Something like `AgendaSynchronizer` would suit better here IMO.
Also you'll need a function here to cancel a sync in case the user logs out.

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
philipplackner: I think it's rather the workers responsibility to delete the events by sync type after the sync has been successful. Otherwise you'll need this wherever you listen to the worker's result

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
philipplackner: If you use async and then await right after, it doesn't do anything. You can just execute the `sync()` function normally in the viewModelScope.launch block. Having multiple launch/async blocks only makes sense if you want to execute multiple suspend functions in parallel which isn't the case here

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
philipplackner: I personally only work with flows and converted the LiveData to a Flow (I think you need a dependency for that). Then did something like this to listen for the result after I called enqueue():

Bildschirmfoto 2023-04-17 um 22 01 45

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
steve1rm: @philipplackner Updated

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsImp.kt
steve1rm: @philipplackner 
 It is identical for `REPLACE`. But for readability reasons it is better to use
     * `CANCEL_AND_REENQUEUE`, because for a reader the difference between `REPLACE` vs `UPDATED`
     * is unclear.

I changed to CANCEL_AND_REENQUEUE

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsWorker.kt
steve1rm: @philipplackner Updated, Can you check as I have changed to using supervisorScope

data/src/main/java/me/androidbox/data/worker_manager/SyncAgendaItemsWorker.kt
steve1rm: @philipplackner Removed the `toList()`

domain/src/main/java/me/androidbox/domain/work_manager/SyncAgendaItems.kt
steve1rm: @philipplackner Adds function to cancel all work. 
```
override suspend fun cancel() {
        workManager.cancelAllWork()
    }
```

PR #47, opened 2023-04-17
[45] open agenda item from notification

steve1rm:
# Pull Request template

###  What does this PR solve?

Open the event screen from the notification using a deeplink 

notifications

### DIFF
--- a/data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
+++ b/data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
@@ -5,7 +5,6 @@ import android.app.PendingIntent
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
-import android.net.Uri
 import androidx.core.app.NotificationCompat
 import androidx.core.app.TaskStackBuilder
 import androidx.core.net.toUri
@@ -13,12 +12,10 @@ import me.androidbox.data.R
 import me.androidbox.data.alarm_manager.AlarmSchedulerImp.Companion.EXTRA_DESCRIPTION
 import me.androidbox.data.alarm_manager.AlarmSchedulerImp.Companion.EXTRA_ID
 import me.androidbox.data.alarm_manager.AlarmSchedulerImp.Companion.EXTRA_TITLE
+import me.androidbox.domain.constant.AgendaDeepLinks.EVENT_DEEPLINK
 
 class AlarmReceiver : BroadcastReceiver() {
 
-    companion object {
-        private const val EVENT_DEEPLINK = "https://androidbox.me/event?agendaId="
-    }
     override fun onReceive(context: Context?, intent: Intent?) {
         if(intent != null && context != null) {
             intent.extras?.let { bundle ->
@@ -39,7 +36,7 @@ class AlarmReceiver : BroadcastReceiver() {
         /** TODO Only Events for the moment, tasks and reminders coming soon... */
         val agendaIntent = Intent(
             Intent.ACTION_VIEW,
-            "$EVENT_DEEPLINK={$agendaId}".toUri())
+            "$EVENT_DEEPLINK".replace("{id}", agendaId).toUri())
 
         val pendingIntent = TaskStackBuilder.create(context).run {
             this.addNextIntentWithParentStack(agendaIntent)
--- a/data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
+++ b/data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
@@ -1,10 +1,14 @@
 package me.androidbox.data.alarm_manager
 
 import android.app.NotificationManager
+import android.app.PendingIntent
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
 import androidx.core.app.NotificationCompat
+import androidx.core.app.TaskStackBuilder
+import androidx.core.net.toUri
 import me.androidbox.data.R
 import me.androidbox.data.alarm_manager.AlarmSchedulerImp.Companion.EXTRA_DESCRIPTION
 import me.androidbox.data.alarm_manager.AlarmSchedulerImp.Companion.EXTRA_ID
@@ -12,6 +16,9 @@ import me.androidbox.data.alarm_manager.AlarmSchedulerImp.Companion.EXTRA_TITLE
 
 class AlarmReceiver : BroadcastReceiver() {
 
+    companion object {
+        private const val EVENT_DEEPLINK = "https://androidbox.me/event?agendaId="
+    }
     override fun onReceive(context: Context?, intent: Intent?) {
         if(intent != null && context != null) {
             intent.extras?.let { bundle ->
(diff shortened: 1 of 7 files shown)


data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
```diff
@@ -25,15 +32,26 @@ class AlarmReceiver : BroadcastReceiver() {
...
     }
-    
+
     private fun createAndShowNotification(context: Context, agendaId: String, title: String, description: String) {
         val notificationManager = getNotificationManager(context)
 
+        /** TODO Only Events for the moment, tasks and reminders coming soon... */
+        val agendaIntent = Intent(
+            Intent.ACTION_VIEW,
+            "$EVENT_DEEPLINK={$agendaId}".toUri())
```
steve1rm: Is this correct in passing the agendID to the deeplink? This is only for Events at the moment

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
```diff
@@ -25,15 +32,26 @@ class AlarmReceiver : BroadcastReceiver() {
...
+
+        val pendingIntent = TaskStackBuilder.create(context).run {
+            this.addNextIntentWithParentStack(agendaIntent)
+            this.getPendingIntent(101, PendingIntent.FLAG_IMMUTABLE or PendingIntent.FLAG_CANCEL_CURRENT)
```
steve1rm: For the request code. Can I use any number for this? I just added 101

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
```diff
@@ -140,7 +141,10 @@ fun NavigationGraph(
                 defaultValue = null
             }, navArgument(MENU_ACTION_TYPE) {
                     type = NavType.StringType
-                })
+                }),
+            deepLinks = listOf(navDeepLink {
+                this.uriPattern = "https://androidbox.me/event{eventId}/{menuActionType}"
```
steve1rm: Is this correct I need to add both the eventId, and menuActionType. The menuActionType was used as part of the dropdownmenu for open/edit/delete.

steve1rm: Another question do I need to append anything to the route: `route = Screen.EventScreen.route`

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
```diff
@@ -1,17 +1,24 @@
...
+import androidx.core.net.toUri
 import me.androidbox.data.R
 import me.androidbox.data.alarm_manager.AlarmSchedulerImp.Companion.EXTRA_DESCRIPTION
 import me.androidbox.data.alarm_manager.AlarmSchedulerImp.Companion.EXTRA_ID
 import me.androidbox.data.alarm_manager.AlarmSchedulerImp.Companion.EXTRA_TITLE
 
 class AlarmReceiver : BroadcastReceiver() {
 
+    companion object {
+        private const val EVENT_DEEPLINK = "https://androidbox.me/event?agendaId="
```
MKiperszmid: I would rename agendaId to **id**

Since agendaId doesnt make much sense

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
MKiperszmid: The event id? Yea its correct

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
MKiperszmid: 1- You are missing a / between the word "event" and the id "eventId"

2- Why do you need to add the menuActionType? Maybe you can make it so the default value is "Open"

3- You shouldnt need to append anything on the route

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
MKiperszmid: Mmmm I think it shouldnt really matter in this case, but i'd still recommend having something like the hashcode

Mostly because I dont know what might happen if you receive 2 notifications from different events. No idea what can happen tho the notification click there

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
MKiperszmid: Also, you could have this on a Constants class
Since you will have to add it to the navgraph also

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
steve1rm: @MKiperszmid Updated to the following to pass in a default OPEN

```
navArgument(MENU_ACTION_TYPE) {
                type = NavType.StringType
                defaultValue = AgendaMenuActionType.OPEN
                })
```

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
steve1rm: @MKiperszmid uses the agendaID.hashcode()

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
steve1rm: @MKiperszmid Adds these to a constants class

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
```diff
@@ -139,8 +141,12 @@ fun NavigationGraph(
                 nullable = true
                 defaultValue = null
             }, navArgument(MENU_ACTION_TYPE) {
-                    type = NavType.StringType
-                })
+                type = NavType.StringType
+                defaultValue = AgendaMenuActionType.OPEN.name
+                }),
```
steve1rm, line 146: @MKiperszmid Is this the correct way to set the default value for the MenuActionType so it will also be open for the deeplink?

data/src/main/java/me/androidbox/data/alarm_manager/AlarmReceiver.kt
```diff
@@ -25,15 +29,26 @@ class AlarmReceiver : BroadcastReceiver() {
...
     }
-    
+
     private fun createAndShowNotification(context: Context, agendaId: String, title: String, description: String) {
         val notificationManager = getNotificationManager(context)
 
+        /** TODO Only Events for the moment, tasks and reminders coming soon... */
+        val agendaIntent = Intent(
+            Intent.ACTION_VIEW,
+            "$EVENT_DEEPLINK".replace("{id}", agendaId).toUri())
```
MKiperszmid, line 42: I would replace agendaId with something like agendaItemId

Since you are refering to the id of an item in the agenda

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
MKiperszmid: Yea

MKiperszmid (approved):

PR #48, opened 2023-04-18
[46] Re-schedule all alarms after rebooting the app 

steve1rm:
# Pull Request template

###  What does this PR solve?

- After rebooting the app all alarms will be scheduled

### DIFF
--- a/data/src/main/java/me/androidbox/data/alarm_manager/AlarmBootReceiver.kt
+++ b/data/src/main/java/me/androidbox/data/alarm_manager/AlarmBootReceiver.kt
@@ -3,7 +3,17 @@ package me.androidbox.data.alarm_manager
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.cancelChildren
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.supervisorScope
+import kotlinx.coroutines.withContext
 import me.androidbox.data.local.agenda.AgendaLocalRepositoryImp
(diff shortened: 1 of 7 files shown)


data/src/main/java/me/androidbox/data/alarm_manager/AlarmBootReceiver.kt
```diff
@@ -0,0 +1,62 @@
...
+                                agendaId = agendaItem.id,
+                                title = agendaItem.title,
+                                description = agendaItem.description,
+                                remindAt = agendaItem.remindAt,
+                                agendaType = agendaItem.agendaType
+                            )
+                        )
+                    }
+                    
+                    coroutineContext.cancelChildren()
```
steve1rm: I don't want to keep this coroutines running as the boardcast has finished. So is it good to cancel like this?

philipplackner: When this job finishes, nothing is running, so this isn't needed

data/src/main/java/me/androidbox/data/local/agenda/AgendaLocalRepositoryImp.kt
```diff
@@ -73,6 +76,30 @@ class AgendaLocalRepositoryImp @Inject constructor(
         eventDao.deleteSyncEventsBySyncType(syncAgendaType)
     }
 
+    override suspend fun fetchAllRemindAtFromEvents(): List<Event> {
+        val eventEntity = eventDao.getAllRemindAt(System.currentTimeMillis())
```
philipplackner, line 80: Should be plural right? Also I find `getAllRemindAt` a bit confusing together with `eventEntity`. The function name sounds like it returns remind at times, but apparently it returns event entities? 🤔

philipplackner (approved):

data/src/main/java/me/androidbox/data/alarm_manager/AlarmBootReceiver.kt
steve1rm: @philipplackner Updated to remove the cancelChildren

PR #49, opened 2023-04-18
[47] sync full agenda

steve1rm:
# Pull Request template

###  What does this PR solve?

- Using work manager to sync the full agenda with local cache

### DIFF
--- a/data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
+++ b/data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
@@ -16,13 +16,13 @@ class FullAgendaSynchronizerImp @Inject constructor(
     private val workManager: WorkManager) : FullAgendaSynchronizer {
 
     companion object {
-        const val THIRTY_MINUTES = 30L
+        const val SYNC_INTERVAL = 30L
         const val FULL_AGENDA_SYNC = "full_agenda_sync"
     }
 
     override suspend fun sync(): UUID {
         val fullSynWorkerRequest = PeriodicWorkRequestBuilder<SyncFullAgendaItemsWorker>(
-            repeatInterval = THIRTY_MINUTES,
+            repeatInterval = SYNC_INTERVAL,
             repeatIntervalTimeUnit = TimeUnit.MINUTES)
             .setConstraints(Constraints(Constraints.Builder().setRequiredNetworkType(NetworkType.CONNECTED).build()))
             .setBackoffCriteria(
--- a/data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
+++ b/data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
@@ -1,14 +1,42 @@
 package me.androidbox.data.worker_manager
 
+import androidx.work.BackoffPolicy
+import androidx.work.Constraints
+import androidx.work.ExistingPeriodicWorkPolicy
+import androidx.work.NetworkType
+import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkManager
+import androidx.work.WorkRequest
 import me.androidbox.domain.work_manager.FullAgendaSynchronizer
+import java.util.UUID
+import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 
 class FullAgendaSynchronizerImp @Inject constructor(
     private val workManager: WorkManager) : FullAgendaSynchronizer {
 
-    override suspend fun sync() {
+    companion object {
+        const val THIRTY_MINUTES = 30L
+        const val FULL_AGENDA_SYNC = "full_agenda_sync"
+    }
 
+    override suspend fun sync(): UUID {
+        val fullSynWorkerRequest = PeriodicWorkRequestBuilder<SyncFullAgendaItemsWorker>(
+            repeatInterval = THIRTY_MINUTES,
+            repeatIntervalTimeUnit = TimeUnit.MINUTES)
+            .setConstraints(Constraints(Constraints.Builder().setRequiredNetworkType(NetworkType.CONNECTED).build()))
+            .setBackoffCriteria(
+                BackoffPolicy.EXPONENTIAL,
+                WorkRequest.MIN_BACKOFF_MILLIS,
+                TimeUnit.MILLISECONDS
+            ).build()
+
+        workManager.enqueueUniquePeriodicWork(
+            FULL_AGENDA_SYNC,
+            ExistingPeriodicWorkPolicy.CANCEL_AND_REENQUEUE,
+            fullSynWorkerRequest)
+
+        return fullSynWorkerRequest.id
     }
 
     override suspend fun cancel() {
--- a/data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
+++ b/data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
@@ -40,6 +40,6 @@ class FullAgendaSynchronizerImp @Inject constructor(
     }
 
     override suspend fun cancel() {
-        workManager.cancelAllWork()
+        workManager.cancelUniqueWork(FULL_AGENDA_SYNC)
     }
 }
--- a/data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
+++ b/data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
@@ -20,7 +20,7 @@ class FullAgendaSynchronizerImp @Inject constructor(
         const val FULL_AGENDA_SYNC = "full_agenda_sync"
     }
 
-    override suspend fun sync(): UUID {
+    override fun sync(): UUID {
         val fullSynWorkerRequest = PeriodicWorkRequestBuilder<SyncFullAgendaItemsWorker>(
             repeatInterval = SYNC_INTERVAL,
             repeatIntervalTimeUnit = TimeUnit.MINUTES)
@@ -39,7 +39,7 @@ class FullAgendaSynchronizerImp @Inject constructor(
         return fullSynWorkerRequest.id
     }
 
-    override suspend fun cancel() {
+    override fun cancel() {
         workManager.cancelUniqueWork(FULL_AGENDA_SYNC)
     }
 }
--- a/data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
+++ b/data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
@@ -1,0 +1,17 @@
+package me.androidbox.data.worker_manager
+
+import androidx.work.WorkManager
+import me.androidbox.domain.work_manager.FullAgendaSynchronizer
+import javax.inject.Inject
+
+class FullAgendaSynchronizerImp @Inject constructor(
+    private val workManager: WorkManager) : FullAgendaSynchronizer {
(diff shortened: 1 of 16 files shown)


presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -40,6 +42,7 @@ class AgendaViewModel @Inject constructor(
         getAuthenticatedUser()
         fetchAgendaItems(ZonedDateTime.now())
         syncAgendaItems()
+        syncFullAgendaItems()
```
steve1rm, line 45: Is this the best place to sync the FullAgenda in the AgendaViewModel?

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
```diff
@@ -0,0 +1,127 @@
...
+
+suspend fun <T, R> List<T>.mapAsync(transformation: suspend (T) -> R): List<R> {
+    return coroutineScope {
+        this@mapAsync.map { item ->
+            async {
+                transformation(item)
+            }
+        }.awaitAll()
+    }
+}
```
steve1rm, line 104: I got this method from the following website. I thought I would try it out. 
https://kt.academy/article/cc-recipes

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
```diff
@@ -0,0 +1,127 @@
...
+
+                reminderDao.insertReminder(reminderEntity)
+            }
+        }
+
+        eventJob.join()
+        taskJob.join()
+        reminderJob.join()
+    }
+}
```
steve1rm, line 94: Is this correct for launching multiple coroutines and then joining all of them?

data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
```diff
@@ -0,0 +1,45 @@
...
+                WorkRequest.MIN_BACKOFF_MILLIS,
+                TimeUnit.MILLISECONDS
+            ).build()
+
+        workManager.enqueueUniquePeriodicWork(
+            FULL_AGENDA_SYNC,
+            ExistingPeriodicWorkPolicy.CANCEL_AND_REENQUEUE,
+            fullSynWorkerRequest)
+
+        return fullSynWorkerRequest.id
```
steve1rm, line 39: Just wondering, what is the purpose of returning a UUID of the worker if I am not using the result anywhere?

data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
```diff
@@ -0,0 +1,45 @@
...
+import me.androidbox.domain.work_manager.FullAgendaSynchronizer
+import java.util.UUID
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+class FullAgendaSynchronizerImp @Inject constructor(
+    private val workManager: WorkManager) : FullAgendaSynchronizer {
+
+    companion object {
+        const val THIRTY_MINUTES = 30L
```
MKiperszmid: I Would rename it to "SYNC_INTERVAL" or similar
What if in the future you change it to 15 min or 1 hour? You need to also rename a variable

data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
MKiperszmid: Returning the ID is used if in the future you want to observe the state

On tasky, it's not really necessary since you won't use it

data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
```diff
@@ -0,0 +1,45 @@
...
+            FULL_AGENDA_SYNC,
+            ExistingPeriodicWorkPolicy.CANCEL_AND_REENQUEUE,
+            fullSynWorkerRequest)
+
+        return fullSynWorkerRequest.id
+    }
+
+    override suspend fun cancel() {
+        workManager.cancelAllWork()
+    }
```
MKiperszmid: What's the use of having this here?

This will also cancel workers that aren't related tot he agenda sync

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
```diff
@@ -0,0 +1,127 @@
...
+            onFailure = {
+                Result.retry()
+            }
+        )
+
+        return result
+    }
+}
+
+private suspend fun insertAllAgendaItems(fullAgendaDto: FullAgendaDto, eventDao: EventDao, taskDao: TaskDao, reminderDao: ReminderDao) {
```
MKiperszmid, line 66: Why is this method outside of the worker?
This can be inside of it, and you'd avoid passing 
```
 eventDao: EventDao, taskDao: TaskDao, reminderDao: ReminderDao
```
Since you'll have those in the worker

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
```diff
@@ -0,0 +1,127 @@
...
+                    eventCreatorId = eventDto.host,
+                    isUserEventCreator = eventDto.isUserEventCreator,
+                    photos = eventDto.photos.map { photoDto ->  photoDto.key },
+                    isGoing = true,
+                    attendees = eventDto.attendees.map { attendeeDto ->
+                        attendeeDto.toAttendee()
+                    }
+                )
+
+                eventDao.insertEvent(eventEntity)
```
MKiperszmid: Instead of inserting one item at a time, you could have an "insertEvents" that accepts a list

So then you'd just:
```
val eventEntities = fullAgendaDto.events.map {it.toEntity()}
eventDao.insertEvents(eventEntities)
```

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
```diff
@@ -0,0 +1,127 @@
...
+                    endDateTime = eventDto.to,
+                    remindAt = eventDto.remindAt,
+                    eventCreatorId = eventDto.host,
+                    isUserEventCreator = eventDto.isUserEventCreator,
+                    photos = eventDto.photos.map { photoDto ->  photoDto.key },
+                    isGoing = true,
+                    attendees = eventDto.attendees.map { attendeeDto ->
+                        attendeeDto.toAttendee()
+                    }
+                )
```
MKiperszmid: Mapping logic shouldn't happen here
I recommend having an EventMapper class

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
MKiperszmid: Yea it's fine

What I like to do, is:

```
(eventJob + taskJob + reminderJob).forEach { it.join() }
```

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
MKiperszmid: No need for this tbh. Since you are already running each stuff in it's own job, inside a supervisorScope
Ex:
```
val eventJob = launch {
```

You are creating a new supervisorScope inside the job
It's really not necessary

If you apply the changes I mentioned above, you can actually remove the mapAsync

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
MKiperszmid: Yea

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
steve1rm: @MKiperszmid I get the following error
`Using 'plus(Job): Job' is an error. Operator '+' on two Job objects is meaningless. Job is a coroutine context element and `+` is a set-sum operator for coroutine contexts. The job to the right of `+` just replaces the job the left of `+``

data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
steve1rm: @MKiperszmid updated

data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
steve1rm: @MKiperszmid Updated

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
steve1rm: @MKiperszmid updated

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
steve1rm: @MKiperszmid updated

data/src/main/java/me/androidbox/data/worker_manager/FullAgendaSynchronizerImp.kt
```diff
@@ -0,0 +1,45 @@
...
+            FULL_AGENDA_SYNC,
+            ExistingPeriodicWorkPolicy.CANCEL_AND_REENQUEUE,
+            fullSynWorkerRequest)
+
+        return fullSynWorkerRequest.id
+    }
+
+    override suspend fun cancel() {
+        workManager.cancelUniqueWork(FULL_AGENDA_SYNC)
+    }
```
philipplackner: As in WhatsApp this and sync don't need to be suspending

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
philipplackner: It doesn't work like @MKiperszmid said. Works only if you're having lists. You'd need to use `listOf(...)` and then you can just call the 3 joins right away

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
philipplackner: You can have such a helper function, but it limits a bit the flexibility since it forces you to use `coroutineScope` and to await directly afterwards. I'd rather return a list of deferreds. If you need to map multiple lists you usually want to call await on all of them after mapping all of them.

philipplackner (approved):

data/src/main/java/me/androidbox/data/worker_manager/SyncFullAgendaItemsWorker.kt
MKiperszmid: 100% right. You would need to have it as a list. My bad!

PR #52, opened 2023-04-21
[51] visitor filter grouping

steve1rm:
# Pull Request template

###  What does this PR solve?

- Updates the visitor section to display going and not going visitors
- Displays the visitor once added by using the dialog box

https://github.com/steve1rm/BusbyTasky/assets/14260802/08355ea2-51e9-47b3-b3aa-854b3f3a4862


component/src/main/java/me/androidbox/component/event/VisitorSection.kt
```diff
@@ -15,34 +17,94 @@ import androidx.compose.ui.tooling.preview.Preview
...
+                        VisitorItem(
+                            visitorInitials = visitor.initials,
+                            visitor = visitor,
+                            isCreator = visitor.isCreator,
+                            onDeleteClicked = visitorDeleteClicked
+                        )
+                    }
+                }
+            }
+        }
```
steve1rm: What I am trying to do here is have `item` which acts like a header for `going` and `not going`
And have 2 `items` blocks that will display the visitors for `going`  or  `not going` or both if `All` is selected
I filter out the what I want display for `going` or `not going` 

`items(visitors.filter { visitorInfo -> visitorInfo.isGoing })`

However, I think it could be inefficient to filter like if this causes multiple  re-composes.

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -229,12 +240,16 @@ class EventViewModel @Inject constructor(
...
+                        onEventScreenEvent(EventScreenEvent.OnAttendeeAdded(attendee))
+                        onEventScreenEvent(EventScreenEvent.OnShowVisitorDialog(false))
+/*
                         _eventScreenState.update { eventScreenState ->
                             eventScreenState.copy(
                                 isEmailVerified = true,
                                 attendees = eventScreenState.attendees + attendee
                             )
                         }
+*/
```
steve1rm: Just a general question I have commented out the code above and calling the `onEventScreenEvent` directly. Which I think keeps all the state updates in a single function `onEventScreenEvent` so there is only one way to update the states.

But as we are in the EventViewModel anywhere it is still worth doing it like this. Or is doing before in the commented code better?

component/src/main/java/me/androidbox/component/event/VisitorSection.kt
```diff
@@ -15,34 +17,94 @@ import androidx.compose.ui.tooling.preview.Preview
...
+import java.util.UUID
+import kotlin.random.Random
 
 @Composable
 fun VisitorSection(
+    visitors: List<VisitorInfo>,
     modifier: Modifier = Modifier,
-    @StringRes titleResId: Int,
-    content: @Composable () -> Unit
+    visitorDeleteClicked: (VisitorInfo) -> Unit,
```
MKiperszmid: Modifier should be after the last mandatory parameter

component/src/main/java/me/androidbox/component/event/VisitorSection.kt
```diff
@@ -15,34 +17,94 @@ import androidx.compose.ui.tooling.preview.Preview
...
+        Spacer(modifier = Modifier.height(16.dp))
 
-        Text(
-            text = stringResource(id = titleResId),
-            fontWeight = FontWeight.Medium,
-            fontSize = 16.sp,
-            color = MaterialTheme.colorScheme.visitorTextFontColor)
+        LazyColumn(modifier = Modifier
+            .fillMaxWidth()
+            .height(300.dp),
```
MKiperszmid: Do you really need this height tho?

component/src/main/java/me/androidbox/component/event/VisitorSection.kt
```diff
@@ -15,34 +17,94 @@ import androidx.compose.ui.tooling.preview.Preview
...
+                    item {
+                        Text(
+                            text = stringResource(id = R.string.going),
+                            fontWeight = FontWeight.Medium,
+                            fontSize = 16.sp,
+                            color = MaterialTheme.colorScheme.visitorTextFontColor
+                        )
+                    }
+
+                    items(visitors.filter { visitorInfo -> visitorInfo.isGoing }) { visitor ->
```
MKiperszmid: Its better to filter on the viewmodel
Or else, you will call this on every recomposition

component/src/main/java/me/androidbox/component/event/VisitorSection.kt
MKiperszmid: Exactly!
Its better to do it on the vm


So then, you would have:

HeaderItem, that displays the selection (All, Going, Not Going), and then a LazyColumn that displays the filtered list

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
MKiperszmid: Im not understanding one thing

After calling verifyVisitorEmail, you are adding the attendee to the list of attendees. But you should only add it when the user confirms about adding it

So the "OnAttendeeAdded" should never happen here
Same with the OnShowVisitorDialog(false)

MKiperszmid (approved):

component/src/main/java/me/androidbox/component/event/VisitorSection.kt
steve1rm: @MKiperszmid 
I was having issues. I think because the outer column can't know the exact height because of the nested lazycolumn. Without this no visitors are shown. Is there a better way to do this?

philipplackner: Yes, I'd put the whole event screen in a single LazyColumn

steve1rm: @philipplackner 
Just wondering how that would work. My EventScreen has something like this below. I am wondering which column would be the best candidate for a LazyColumn. And just wondering what advantages of having a LazyColumn over a conventional Column? 
```
 Scaffold(
        modifier = modifier,
        topBar = {
            AgendaDetailTopBar()
        },
        content = {
            Column() {

                Column () {
                      AgendaHeader()

                      PhotoPicker()

                      AgendaDuration()

                      AgendaDropDownMenu()

                      AlarmReminder()
                    
                      VisitorFilter()

                      VisitorSection()

                      AgendaAction()
                }
       }
}
```
Would each composable be wrapped inside an `item` block?

philipplackner: Yes correct, you'd use individual item blocks. Just for the attendees list make sure to wrap these in `items` instead, otherwise the whole list will be treated as one composable and you won't benefit from the lazy loading behavior.

data/src/main/java/me/androidbox/data/local/entity/EventEntity.kt
```diff
@@ -16,7 +16,7 @@ data class EventEntity(
     val remindAt: Long,
     val eventCreatorId: String,
     val isUserEventCreator: Boolean,
-    val isGoing: Boolean,
+    val host: String,
```
steve1rm, line 19: I removed the isGoing as I noticed in the /agenda, create /event, and get /event there is no `isGoing`. But here is a `host`. So I use this to see if the user is a creator or not.

MKiperszmid: Sounds good

data/src/main/java/me/androidbox/data/worker_manager/UploadEventWorker.kt
```diff
@@ -43,15 +43,21 @@ class UploadEventWorker @AssistedInject constructor(
                 createPhotoMultipart.createMultipartPhotos(listOfPhoto.toList())
             } ?: listOf()
 
-            if(runAttemptCount > RETRY_COUNT) {
+            if(runAttemptCount < RETRY_COUNT) {
                 Result.failure()
```
MKiperszmid: This is incorrect

if runAttemptCount > Retry Count, it should return Failure

Else, return Result.retry

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
```diff
@@ -11,11 +11,15 @@ import java.time.ZonedDateTime
...
     val listOfPhotoUri: List<String> = mutableStateListOf<String>(),
-    val selectedVisitorType: VisitorFilterType = VisitorFilterType.ALL,
     val selectedVisitor: VisitorInfo? = null,
     val visitorEmail: String = "",
     val shouldShowVisitorDialog: Boolean = false,
+    val selectedVisitorFilterType: VisitorFilterType = VisitorFilterType.ALL,
+    val filteredVisitorsGoing: List<Attendee> = mutableStateListOf(),
+    val filteredVisitorsNotGoing: List<Attendee> = mutableStateListOf(),
     val isEmailVerified: Boolean = true,
+    val isAlreadyAdded: Boolean = false,
```
MKiperszmid, line 22: Why do you need this?

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
```diff
@@ -27,9 +31,11 @@ data class EventScreenState(
     val isStartDateTime: Boolean = false,
     val shouldOpenDropdown: Boolean = false,
     val alarmReminderItem: AlarmReminderItem = AlarmReminderItem.TEN_MINUTES,
-    val attendees: List<Attendee> = listOf(),
+    val attendees: List<Attendee> = mutableStateListOf(),
```
MKiperszmid, line 35: Why mutablelist? You could have a immutablelist and just override it

If it's mutable, then it can be modified by the ui

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
```diff
@@ -11,11 +11,15 @@ import java.time.ZonedDateTime
 
 data class EventScreenState(
     val listOfPhotoUri: List<String> = mutableStateListOf<String>(),
-    val selectedVisitorType: VisitorFilterType = VisitorFilterType.ALL,
     val selectedVisitor: VisitorInfo? = null,
     val visitorEmail: String = "",
     val shouldShowVisitorDialog: Boolean = false,
+    val selectedVisitorFilterType: VisitorFilterType = VisitorFilterType.ALL,
+    val filteredVisitorsGoing: List<Attendee> = mutableStateListOf(),
+    val filteredVisitorsNotGoing: List<Attendee> = mutableStateListOf(),
```
MKiperszmid, line 20: Same mutableList here as in the other comment I made

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -109,10 +101,18 @@ class EventViewModel @Inject constructor(
...
+                val attendee = eventScreenState.value.attendees.firstOrNull { attendee ->
+                    attendee.userId == eventScreenEvent.userId
+                }
+
+                if(attendee != null) {
+                    _eventScreenState.update { eventScreenState ->
+                        eventScreenState.copy(
+                            attendees = eventScreenState.attendees - attendee,
+                            filteredVisitorsGoing = eventScreenState.attendees.filter { it.isGoing },
+                            filteredVisitorsNotGoing = eventScreenState.attendees.filter { !it.isGoing }
```
MKiperszmid: This could be a method, as to not have to do it over and over again

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -177,7 +177,9 @@ class EventViewModel @Inject constructor(
             is EventScreenEvent.OnAttendeeAdded -> {
                 _eventScreenState.update { eventScreenState ->
                     eventScreenState.copy(
-                        attendees = eventScreenState.attendees + eventScreenEvent.attendee
+                        attendees = eventScreenState.attendees + eventScreenEvent.attendee.copy(isGoing = true),
+                        filteredVisitorsGoing = eventScreenState.attendees.filter { it.isGoing },
+                        filteredVisitorsNotGoing = eventScreenState.attendees.filter { !it.isGoing }
                     )
                 }
             }
```
MKiperszmid: You are going to call this only in the ViewModel. No need for an event. You can just have a function for it, and call it on the verifyVisitorEmail method

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -232,12 +264,10 @@ class EventViewModel @Inject constructor(
...
 
-                        _eventScreenState.update { eventScreenState ->
-                            eventScreenState.copy(
-                                isEmailVerified = true,
-                                attendees = eventScreenState.attendees + attendee
-                            )
-                        }
+                        onEventScreenEvent(EventScreenEvent.OnAttendeeAdded(attendee))
+                        onEventScreenEvent(EventScreenEvent.OnVerifyingVisitorEmail(false))
+                        onEventScreenEvent(EventScreenEvent.OnShowVisitorDialog(false))
```
MKiperszmid, line 334: If from the viewmodel you need to call an event, that´s not being called on the UI, then no need to create it as an event
It can just be a private fun

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
steve1rm: @MKiperszmid I'm checking if the user has already added a duplicate attendee

component/src/main/java/me/androidbox/component/agenda/AddVisitorDialog.kt
```diff
@@ -73,11 +76,18 @@ fun AddVisitorDialog(
                 }
             )
 
-            if(!isEmailVerified) {
+            if(!isEmailVerified or isAlreadyAdded) {
```
philipplackner: In general, I'd recommend to use the real logical operators ( `||`) for this. For booleans it doesn't make a difference, but usually the bitwise operators are only used for integers and to combine flags for example for pending intents

domain/src/main/java/me/androidbox/domain/util.kt
```diff
@@ -0,0 +1,14 @@
...
+fun String.toInitials(): String {
+    val listOfName = this.trim().split(" ")
+    var initials = ""
+
+    listOfName.forEach { name ->
+        initials += name.first()
+    }
+
+    return initials.uppercase(Locale.getDefault())
+}
```
philipplackner, line 14: Some questions:
1. What if the full name consists of first, middle and last name? I think the initials should then be from first and last name ignoring the middle name.
2. What if the user only provides their first name as the full name? Then I think it should show the first 2 chars
3. What if there are 2 spaces in the name? Not 100% what `split` would do then but it might result in an empty string and crash here

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreenState.kt
philipplackner: But why do you need this as state? You can simply check this by searching for the added attendee ID in the existing attendees list

philipplackner (approved):

domain/src/main/java/me/androidbox/domain/util.kt
steve1rm: @philipplackner I will write some test cases for those scenarios

component/src/main/java/me/androidbox/component/agenda/AddVisitorDialog.kt
steve1rm: @philipplackner Updated

component/src/main/java/me/androidbox/component/agenda/AddVisitorDialog.kt
```diff
@@ -73,11 +76,18 @@ fun AddVisitorDialog(
...
             )
 
-            if(!isEmailVerified) {
+            if(!isEmailVerified || isAlreadyAdded) {
+                val errorMessage = if(!isEmailVerified) {
+                    stringResource(id = R.string.email_verify_failed)
+                }
+                else {
+                    stringResource(R.string.visitor_already_added)
+                }
```
MKiperszmid: I would use a **when** here 

```
val errorMessage = when {
    !isEmailVerified -> stringResource(id = R.string.email_verify_failed)
    isAlreadyAdded -> stringResource(R.string.visitor_already_added)
    else -> null
}

if (errorMessage != null) {
// Show the text
}
```
(I'm on phone rn so it might not compile 100%)


Another  approach could be to directly pass the error message to the composable, so you just do the calculation on the viewmodel

data/src/main/java/me/androidbox/data/remote/model/response/AttendeeDto.kt
```diff
@@ -4,7 +4,7 @@ data class AttendeeDto(
     val email: String,
     val fullName: String,
     val userId: String,
-  //  val eventId: String,
-  //  val isGoing: Boolean,
-  //  val remindAt: Long
+    val eventId: String = "",
+    val isGoing: Boolean = false,
+    val remindAt: Long = 0L
```
MKiperszmid, line 9: Why the default values?

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -87,110 +81,219 @@ fun EventScreen(
...
+                                fontWeight = FontWeight.Medium,
+                                fontSize = 16.sp,
+                                color = MaterialTheme.colorScheme.visitorTextFontColor
+                            )
+                            Spacer(modifier = Modifier.height(16.dp))
+
                         }
-                    )
-                    Spacer(modifier = modifier.height(26.dp))
+                        if(eventScreenState.filteredVisitorsGoing.isNotEmpty()) {
```
MKiperszmid: No need for this check
Since the "items" already does it internally

If the list is empty, it won't display any VisitorItem

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -87,110 +81,219 @@ fun EventScreen(
...
+                            items(eventScreenState.filteredVisitorsGoing) { attendee ->
+                                VisitorItem(
+                                    initials = attendee.fullName.toInitials(),
```
MKiperszmid: Here you are calculating .toIntials on every recomposition

This app won't have much performance issues, but in the future, you could have it calculated on the ViewModel just once, and use that list

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -87,110 +81,219 @@ fun EventScreen(
...
+                                fontWeight = FontWeight.Medium,
+                                fontSize = 16.sp,
+                                color = MaterialTheme.colorScheme.visitorTextFontColor
+                            )
+                            Spacer(modifier = Modifier.height(16.dp))
                         }
-                    )
 
-                    Spacer(modifier = modifier.height(26.dp))
+                        if(eventScreenState.filteredVisitorsNotGoing.isNotEmpty()) {
```
MKiperszmid: Applies the same comments as mentioned in the "Going" section

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -87,110 +81,219 @@ fun EventScreen(
...
+                                 * isGoing = true ==> LEAVE_EVENT
+                                 * isGoing = false ==> JOIN_EVENT
+                                 * */
 
-                                }
-                                AgendaActionType.LEAVE_EVENT -> {
+                                /* Find the attendee and check their status - TODO Consider a helper method for this */
+                                val isAttendeeGoing = eventScreenState.attendees.firstOrNull { attendee ->
+                                    attendee.userId == eventScreenState.currentLoggedInUserId
+                                }?.isGoing ?: false
```
MKiperszmid, line 286: I would have this on the viewmodel. Try to keep the UI as dumb as possible, and not make much calculations

Again, this won't take much performance, but if in the future you add something complex, it's better for you to get used to adding it directly on the ViewModel
This is due to recomposition

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -87,110 +81,219 @@ fun EventScreen(
...
+
+                                if(isAttendeeGoing)
+                                    AgendaActionType.LEAVE_EVENT
+                                else
+                                    AgendaActionType.JOIN_EVENT
```
MKiperszmid: I would either add the {}, or make it all in one line

It's a bit easier to read (+ it's a best practice on Android's guidelines https://developer.android.com/kotlin/style-guide#formatting )

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
```diff
@@ -241,9 +344,24 @@ fun EventScreen(
             },
             isValidInput = false,
             onAddButtonClicked = { visitorEmail ->
-                eventScreenEvent(EventScreenEvent.CheckVisitorExists(visitorEmail))
+
+                /** TODO Remove this logic for checking attendees already added to the viewModel */
```
MKiperszmid, line 368: I was going to say exactly this

Just have an "AddVisitor" and have the ViewModel do all the checks for if the visitor was already added or not, and if it's a duplicate, etc

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -85,6 +83,13 @@ class EventViewModel @Inject constructor(
                 }
             }
         }
+
+        preferenceRepository.retrieveCurrentUserOrNull()?.let { authenticatedUser ->
+        _eventScreenState.update { eventScreenState ->
+                eventScreenState.copy(
+                    currentLoggedInUserId = authenticatedUser.userId)
```
MKiperszmid, line 91: The formatting here is weird
It should either be:
```
 eventScreenState.copy(currentLoggedInUserId = authenticatedUser.userId)
```
or
```
 eventScreenState.copy(
                    currentLoggedInUserId = authenticatedUser.userId
) // I'm on phone so I can't remember if the ) should be at the start, or more in the middle. But Android Studio should automatically do it
```

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -189,10 +195,33 @@ class EventViewModel @Inject constructor(
...
+                        filteredVisitorsNotGoing = eventScreenState.attendees.filter { !it.isGoing }
                     )
                 }
             }
+            is EventScreenEvent.OnAttendeeStatusUpdate -> {
+                val index = eventScreenState.value.attendees.indexOfFirst { attendee ->
+                    attendee.userId == eventScreenState.value.currentLoggedInUserId
+                }
+
+                if(index > -1) {
```
MKiperszmid: Usually people use 
```
if (index != -1)
```

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -212,6 +241,20 @@ class EventViewModel @Inject constructor(
...
+                    )
+                }
+            }
+            is EventScreenEvent.OnVerifyingVisitorEmail -> {
+                _eventScreenState.update { eventScreenState ->
+                    eventScreenState.copy(
+                        isVerifyingVisitorEmail = eventScreenEvent.isVerifyingVisitorEmail
+                    )
+                }
+            }
```
MKiperszmid, line 257: This could all be just one method that get's called in the "AddVisitor"

ex:

fun validateVisitor()

You validate if the user is added or duplicate or whatever, and you can just return one error message if necessary

So you can avoid the when/if expression in the first comment I made

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -222,6 +265,21 @@ class EventViewModel @Inject constructor(
...
+                _eventScreenState.update {  eventScreenState ->
+                    eventScreenState.copy(
+                        selectedVisitorFilterType = eventScreenEvent.visitorFilterType)
+                }
+            }
+            EventScreenEvent.LoadVisitors -> {
+                _eventScreenState.update { eventScreenState ->
+                    eventScreenState.copy(
+                        filteredVisitorsGoing = eventScreenState.attendees.filter { it.isGoing },
+                        filteredVisitorsNotGoing = eventScreenState.attendees.filter { !it.isGoing }
```
MKiperszmid: You called this a couple of times

Maybe have a method and reuse it

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -235,6 +293,7 @@ class EventViewModel @Inject constructor(
 
     private fun verifyVisitorEmail(visitorEmail: String) {
         viewModelScope.launch {
+            onEventScreenEvent(EventScreenEvent.OnVerifyingVisitorEmail(true))
```
MKiperszmid, line 314: I noticed you create a ton of events that aren't really necessary

This "OnVerifyingVisitorEmail" is a method that's only being used on this verifyVisitorEmail method

So you can remove that event, and replace it with the state updates 
```
_eventScreenState.update { eventScreenState ->
                    eventScreenState.copy(isVerifyingVisitorEmail = true)
                }
```
and
```
_eventScreenState.update { eventScreenState ->
                    eventScreenState.copy(isVerifyingVisitorEmail = false)
                }
```

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
steve1rm: @MKiperszmid Understand, will move this to the ViewModel

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
steve1rm: @MKiperszmid Updated

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
steve1rm: @MKiperszmid Updated

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
steve1rm: @MKiperszmid Will refactor this into a single method

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
steve1rm: @MKiperszmid Will update this

data/src/main/java/me/androidbox/data/remote/model/response/AttendeeDto.kt
steve1rm: @MKiperszmid The default values was because the endpoint doesn't send them so they are optional.

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
steve1rm: @MKiperszmid Updated

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
steve1rm: @MKiperszmid Updated

presentation/src/main/java/me/androidbox/presentation/event/screen/EventScreen.kt
steve1rm: @MKiperszmid Updated

component/src/main/java/me/androidbox/component/agenda/AddVisitorDialog.kt
steve1rm: @MKiperszmid I will check this

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
steve1rm: updated

component/src/main/java/me/androidbox/component/agenda/AddVisitorDialog.kt
steve1rm: Updated

PR #54, opened 2023-04-25
[53] logout

steve1rm:
# Pull Request template

###  What does this PR solve?

Opens the dropdown menu and allows the user to logout

logout

### DIFF
--- a/component/src/main/java/me/androidbox/component/agenda/AgendaHomeTopBar.kt
+++ b/component/src/main/java/me/androidbox/component/agenda/AgendaHomeTopBar.kt
@@ -32,7 +32,7 @@ fun AgendaTopBar(
     onProfileButtonClicked: () -> Unit,
     onDateClicked: () -> Unit,
     modifier: Modifier = Modifier,
-    dropDropMenuScope: @Composable () -> Unit
+    dropDownMenu: @Composable () -> Unit
 ) {
 
     Row(
@@ -68,7 +68,7 @@ fun AgendaTopBar(
                 )
             }
 
-            dropDropMenuScope()
+            dropDownMenu()
         }
     }
 }
@@ -86,7 +86,7 @@ fun PreviewAgendaTopBar() {
             initials = "SM",
             onDateClicked = {},
             onProfileButtonClicked = {},
-            dropDropMenuScope = {}
+            dropDownMenu = {}
         )
     }
 }
--- a/component/src/main/java/me/androidbox/component/agenda/AgendaHomeTopBar.kt
+++ b/component/src/main/java/me/androidbox/component/agenda/AgendaHomeTopBar.kt
@@ -2,7 +2,13 @@ package me.androidbox.component.agenda
 
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
-import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.material3.Button
 import androidx.compose.material3.ButtonDefaults
 import androidx.compose.material3.MaterialTheme
@@ -26,6 +32,7 @@ fun AgendaTopBar(
     onProfileButtonClicked: () -> Unit,
     onDateClicked: () -> Unit,
     modifier: Modifier = Modifier,
+    dropDropMenuScope: @Composable () -> Unit
 ) {
 
     Row(
@@ -43,21 +50,25 @@ fun AgendaTopBar(
             color = MaterialTheme.colorScheme.topbarFontColor
         )
 
-        Button(
-            modifier = Modifier.size(34.dp),
-            contentPadding = PaddingValues(0.dp),
-            colors = ButtonDefaults.buttonColors(
-                containerColor = MaterialTheme.colorScheme.topbarButtonBackgroundColor,
-                contentColor = MaterialTheme.colorScheme.topbarFontColor),
-            onClick = {
-                onProfileButtonClicked()
-            }) {
-            Text(
-                maxLines = 1,
-                text = initials,
-                fontSize = 13.sp,
-                fontWeight = FontWeight.SemiBold,
-            )
+        Box {
+            Button(
+                modifier = Modifier.size(34.dp),
+                contentPadding = PaddingValues(0.dp),
+                colors = ButtonDefaults.buttonColors(
+                    containerColor = MaterialTheme.colorScheme.topbarButtonBackgroundColor,
+                    contentColor = MaterialTheme.colorScheme.topbarFontColor),
+                onClick = {
+                    onProfileButtonClicked()
+                }) {
+                Text(
+                    maxLines = 1,
+                    text = initials,
+                    fontSize = 13.sp,
+                    fontWeight = FontWeight.SemiBold,
+                )
+            }
+
+            dropDropMenuScope()
         }
     }
 }
@@ -75,6 +86,7 @@ fun PreviewAgendaTopBar() {
             initials = "SM",
             onDateClicked = {},
             onProfileButtonClicked = {},
+            dropDropMenuScope = {}
         )
     }
 }
--- a/domain/src/main/java/me/androidbox/domain/authentication/usecase/imp/LogoutUseCaseImp.kt
+++ b/domain/src/main/java/me/androidbox/domain/authentication/usecase/imp/LogoutUseCaseImp.kt
@@ -1,0 +1,14 @@
+package me.androidbox.domain.authentication.usecase.imp
+
+import me.androidbox.domain.authentication.ResponseState
+import me.androidbox.domain.authentication.usecase.AuthenticateUserUseCase
+import me.androidbox.domain.authentication.usecase.LogoutUseCase
+import javax.inject.Inject
+
+class LogoutUseCaseImp @Inject constructor(
+    private val authenticateUserUseCase: AuthenticateUserUseCase
+) : LogoutUseCase {
+    override suspend fun execute(): ResponseState<Unit> {
+        return authenticateUserUseCase.execute()
+    }
+}
--- a/presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
+++ b/presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
@@ -64,7 +64,7 @@ fun AgendaScreen(
                     onDateClicked = {
                         calendarState.show()
                     },
-                    dropDropMenuScope = {
+                    dropDownMenu = {
                         AgendaDropDownMenu(
                             modifier = Modifier
                                 .background(color = MaterialTheme.colorScheme.dropDownMenuBackgroundColor)
--- a/presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
+++ b/presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
@@ -1,6 +1,5 @@
 package me.androidbox.presentation.agenda.screen
 
-import android.util.Log
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.*
 import androidx.compose.foundation.lazy.LazyColumn
@@ -14,7 +13,6 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.maxkeppeker.sheets.core.models.base.rememberUseCaseState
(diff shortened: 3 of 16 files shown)


presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
```diff
@@ -43,28 +44,47 @@ fun AgendaScreen(
...
+                    },
+                    onDateClicked = {
+                        calendarState.show()
+                    },
+                )
+
+                AgendaDropDownMenu(
+                    modifier = Modifier
+                        .background(color = MaterialTheme.colorScheme.dropDownMenuBackgroundColor)
+                        .align(Alignment.End),
```
steve1rm: I am trying to get the Dropdownmenu to display at the end below the profile avatar. But for some reason its always displays at the start. I have posted a screenshot in the description. What is the best way to position at the end below the avatar?

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
```diff
@@ -131,6 +131,10 @@ fun NavigationGraph(
                     AgendaType.TASK -> TODO()
                     AgendaType.REMINDER -> TODO()
                 }
+            },
+            onLogout = {
+                navHostController.popBackStack()
+                navHostController.navigate(Screen.LoginScreen.route)
```
steve1rm: Will this pop all the screens. Or just the current screen which is the AgendaAcreen?

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -125,6 +127,25 @@ class AgendaViewModel @Inject constructor(
...
+                    )
+                }
+            }
+        }
+    }
+
+    private fun logoutCurrentUser() {
+        viewModelScope.launch {
+            logoutUseCase.execute()
+            preferenceRepository.deleteCurrentUser()
```
steve1rm: Do we need to do anything with the database tables?

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
MKiperszmid: Have you tried adding a fillMaxWidth to the AgendaDropDownMenu´s modifier?

If that doesn´t work, you can add the AgendaDropDownMenu inside the profile picture composable

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
```diff
@@ -43,28 +44,47 @@ fun AgendaScreen(
...
+                    shouldOpenDropdown = agendaScreenState.shouldOpenLogoutDropDownMenu,
+                    onCloseDropdown = {
+                        agendaScreenEvent(
+                            AgendaScreenEvent.OnChangedShowDropdownStatus(shouldOpen = false)
+                        )
+                    },
+                    listOfMenuItemId = listOf(me.androidbox.presentation.R.string.logout),
```
MKiperszmid: Why do you need a list tho?
Can´t it just be one id?

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
MKiperszmid: Delete them

What if a 2nd user logs in? Then that user will be able to see the previous user´s tables

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
MKiperszmid: popBackStack only pops the last screen

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
steve1rm: @MKiperszmid Having the fillMaxWidth doesn't work. Do you have anymore ideas?

logout

domain/src/main/java/me/androidbox/domain/authentication/usecase/imp/LogoutUseCaseImp.kt
```diff
@@ -0,0 +1,14 @@
...
+import me.androidbox.domain.authentication.usecase.LogoutUseCase
+import javax.inject.Inject
+
+class LogoutUseCaseImp @Inject constructor(
+    private val authenticateUserUseCase: AuthenticateUserUseCase
+) : LogoutUseCase {
+    override suspend fun execute(): ResponseState<Unit> {
+        return authenticateUserUseCase.execute()
+    }
+}
```
philipplackner, line 14: What is the point of this empty use case executing another use case?

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
philipplackner: Also I'd have some other kind of identifier for the items, otherwise you'll always have to remember the exact string (which can change)

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
philipplackner: Which composable is this drop down anchored to? Usually if you put it in a Box together with another composable it will show below that other composable. If the anchor fills the whole width, the drop down might show where you don't want it, so you can try making the anchor (in this case the profile pic) only fill the space it needs

philipplackner (approved):

domain/src/main/java/me/androidbox/domain/authentication/usecase/imp/LogoutUseCaseImp.kt
steve1rm: @philipplackner Will refactor this. I guess I am rushing to finish this. Will do this at a later time.

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
steve1rm: I have anchored the dropdownmenu below the whole `AgendaTopBar` using the `Box`. I understand what you mean as the `AgendaTopbar` takes up the whole width of the screen. As in the image below. So the dropdownMenu might display on the left. However, is there another way to position the dropdownmenu on the right without moving the dropdownmenu into the `AgendaTopBar`

logout-box

steve1rm: @philipplackner @MKiperszmid 
Uses a composable lambda that the dropdownmenu will be passed to. The result is correct now.

logoutrightside

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
steve1rm: @philipplackner @MKiperszmid Is this the behavior that I want? Or should I be popping all screens?

component/src/main/java/me/androidbox/component/agenda/AgendaHomeTopBar.kt
```diff
@@ -26,6 +32,7 @@ fun AgendaTopBar(
     onProfileButtonClicked: () -> Unit,
     onDateClicked: () -> Unit,
     modifier: Modifier = Modifier,
+    dropDropMenuScope: @Composable () -> Unit
```
steve1rm: @philipplackner Is this the correct way to do this. Any suggestions about the naming?

presentation/src/main/java/me/androidbox/presentation/navigation/NavigationGraph.kt
philipplackner: Pop all screens, otherwise the user can press back on login to get back to an agenda screen

component/src/main/java/me/androidbox/component/agenda/AgendaHomeTopBar.kt
philipplackner: Yes, but the name is a bit weird, why `dropDrop`? What about just `dropDownMenu`?

philipplackner (approved):

component/src/main/java/me/androidbox/component/agenda/AgendaHomeTopBar.kt
steve1rm: @philipplackner that was a typo error :blush:

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
```diff
@@ -43,28 +42,52 @@ fun AgendaScreen(
...
     onSelectedEditAgendaItemClicked: (id: String, AgendaType, agendaMenuActionType: AgendaMenuActionType) -> Unit,
     onSelectedAgendaItem: (agendaType: Int) -> Unit, /* TODO Check where this is being used */
+    onLogout: () -> Unit,
     modifier: Modifier = Modifier) {
 
     val calendarState = rememberUseCaseState()
 
+    LaunchedEffect(key1 = agendaScreenState.deletedCacheCompleted) {
+        onLogout()
+    }
```
steve1rm, line 54: @philipplackner @MKiperszmid I have added a LaunchedEffect here. The reason is I want to only logout once the deleted cache has been completed successfully.  So the `deletedCacheCompleted` will be set to true in the AgendaViewModel after the cache is deleted. Does this look ok to you?

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -125,6 +127,53 @@ class AgendaViewModel @Inject constructor(
...
+        }
+    }
+
+    private fun logoutCurrentUser() {
+        viewModelScope.launch {
+            when(logoutUseCase.execute()) {
+                ResponseState.Loading -> Unit /* TODO Show loading */
+
+                is ResponseState.Success -> {
+                    preferenceRepository.deleteCurrentUser()
```
steve1rm, line 151: @philipplackner @MKiperszmid I didn't launch this in a separate coroutine as is not a suspended function.

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -125,6 +127,53 @@ class AgendaViewModel @Inject constructor(
...
+
+                    listOf(eventJob, taskJob, reminderJob).forEach { job ->
+                        job.join()
+                    }
+
+                    _agendaScreenState.update { agendaScreenState ->
+                        agendaScreenState.copy(
+                            deletedCacheCompleted = true
+                        )
+                    }
```
steve1rm, line 171: @philipplackner @MKiperszmid This is the state that is set to true after the cache as been deleted. That will trigger the `logout()` in the LaunchedEffect.  Does this look ok?

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -125,6 +127,53 @@ class AgendaViewModel @Inject constructor(
...
+
+    private fun logoutCurrentUser() {
+        viewModelScope.launch {
+            when(logoutUseCase.execute()) {
+                ResponseState.Loading -> Unit /* TODO Show loading */
+
+                is ResponseState.Success -> {
+                    preferenceRepository.deleteCurrentUser()
+
+                    val eventJob = viewModelScope.launch {
```
MKiperszmid, line 153: Room database has a "deleteAllTables" method

You can just call that

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
MKiperszmid: Gotcha

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
MKiperszmid: Yes

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
MKiperszmid: Yea

But i would add an "if deleted cache is true"

You dont want to trigger this LaunchedEffect on the first recomposition

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/agenda/screen/AgendaScreen.kt
steve1rm: @MKiperszmid The `deletedCacheCompleted` is set to false initially in my state class. However, how would I prevent this from triggering in the first composition.

MKiperszmid: Replied on whatsapp

But:

It will always run on the first composition
Not possible to avoid it
Thats why u need an if check

PR #55, opened 2023-04-28
[54] Deleting events from the remote endpoint

steve1rm:
# Pull Request template

###  What does this PR solve?

Deleting events using the /event?eventId endpoint

Offline mode
1) Delete from the room DB
2) Insert the event ID into the syncTable 
3) Fetch from DB (Will contain fetched items that were inserted from the EP)
4) When back online run sync worker /synAgenda

Online mode
1) Delete from the room DB
2) Delete from the /event EP
3) Fetch from DB with events inserted from EP

The only difference is inserting to the sync table when offline, to be sync'ed when online again.

**NOTE: I haven't completed checking for network connection**

### DIFF
--- a/data/src/main/java/me/androidbox/data/remote/event/DeleteEventWithIdRemoteUseCaseImp.kt
+++ b/data/src/main/java/me/androidbox/data/remote/event/DeleteEventWithIdRemoteUseCaseImp.kt
@@ -1,0 +1,27 @@
+package me.androidbox.data.remote.event
+
+import me.androidbox.data.remote.network.event.EventService
+import me.androidbox.data.remote.util.CheckResult.checkResult
+import me.androidbox.domain.authentication.ResponseState
+import me.androidbox.domain.event.usecase.DeleteEventWithIdRemoteUseCase
+import javax.inject.Inject
+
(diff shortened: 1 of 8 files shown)


data/src/main/java/me/androidbox/data/remote/event/DeleteEventWithIdRemoteUseCaseImp.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import me.androidbox.data.remote.network.event.EventService
+import me.androidbox.data.remote.util.CheckResult.checkResult
+import me.androidbox.domain.authentication.ResponseState
+import me.androidbox.domain.event.usecase.DeleteEventWithIdRemoteUseCase
+import javax.inject.Inject
+
+class DeleteEventWithIdRemoteUseCaseImp @Inject constructor(
+    private val eventService: EventService
+) : DeleteEventWithIdRemoteUseCase {
```
philipplackner, line 11: This use case is in your data layer, but should go in domain. In that case, you can't use the data related dependencies though.
The purpose of a use case is to only contain business logic, nothing related to any SDK or data related things.

philipplackner (approved):

PR #56, opened 2023-04-30
[55] Uses aganda calendar day selector

steve1rm:
# Pull Request template

###  What does this PR solve?

- Uses the agenda calendar day picker to fetch the agendas for that day

agenda-day-picker

### DIFF
--- a/presentation/src/main/AndroidManifest.xml
+++ b/presentation/src/main/AndroidManifest.xml
@@ -6,7 +6,9 @@
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
-
+    <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM"/>
+    <uses-permission android:name="android.permission.USE_EXACT_ALARM"/>
+    
     <application
         android:name=".BusbyTaskyApplication"
         android:allowBackup="true"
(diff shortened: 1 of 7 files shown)


presentation/src/main/AndroidManifest.xml
```diff
@@ -6,7 +6,9 @@
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
-
+    <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM"/>
+    <uses-permission android:name="android.permission.USE_EXACT_ALARM"/>
```
steve1rm, line 10: I was getting a cash so I had to add these permissions. Is this correct?
``` Caller me.androidbox.presentation needs to hold android.permission.SCHEDULE_EXACT_ALARM or android.permission.USE_EXACT_ALARM to set exact alarms```

philipplackner: Yes, `USE_EXACT_ALARM` is needed here

philipplackner (approved):

presentation/src/main/AndroidManifest.xml
MKiperszmid: Yes. Also im pretty sure that having UseExactAlarm, you need to ask for that permission at runtime also

For emulators and so on, its not needed
But necessary if you plan to upload to playstore

MKiperszmid (approved):

PR #60, opened 2023-05-02
[59] pull to refresh

steve1rm:
# Pull Request template

###  What does this PR solve?

- Uses swipe to refresh the agenda items

https://user-images.githubusercontent.com/14260802/235571610-49767815-6fb9-48be-8757-fa8db8439192.mp4

I noticed that this is now available in the compose material library. Will migrate later using this guide:
https://google.github.io/accompanist/swiperefresh/

### DIFF
--- a/presentation/build.gradle.kts
+++ b/presentation/build.gradle.kts
@@ -97,7 +97,7 @@ dependencies {
     implementation("io.coil-kt:coil-compose:1.4.0")
     implementation("androidx.work:work-runtime-ktx:2.8.0")
     implementation("androidx.hilt:hilt-work:1.0.0")
-
+    implementation("com.google.accompanist:accompanist-swiperefresh:0.27.0")
     debugImplementation(tests.ui.tooling)
     debugImplementation(tests.ui.test.manifest)
 
--- a/presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
+++ b/presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
@@ -62,7 +62,7 @@ class AgendaViewModel @Inject constructor(
             .minusSeconds(1L)
     }
 
-    fun fetchAgendaItems(agendaDate: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault())) {
+    fun fetchAgendaItems(agendaDate: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault()), isSwipeToRefresh: Boolean = false) {
         agendaJob?.cancel()
 
         agendaJob = viewModelScope.launch {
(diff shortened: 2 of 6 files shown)


presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -62,19 +62,30 @@ class AgendaViewModel @Inject constructor(
             .minusSeconds(1L)
     }
 
-    fun fetchAgendaItems(agendaDate: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault())) {
+    fun fetchAgendaItems(agendaDate: ZonedDateTime = ZonedDateTime.now(ZoneId.systemDefault()), isSwipeToRefresh: Boolean = false) {
```
steve1rm, line 65: I added a flag here as I only want to show the swipeRefresh progress indicator only when the user has swiped to refresh. Otherwise this will show each time the user fetches new agenda items i.e. adding, changing dates, etc. Is this ok?

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -62,19 +62,30 @@ class AgendaViewModel @Inject constructor(
...
                     when(responseState) {
                         ResponseState.Loading -> {
-                            /* TODO Show some form of loading */
+                            if(isSwipeToRefresh) {
+                                _agendaScreenState.update { agendaScreenState ->
+                                    agendaScreenState.copy(
+                                        isRefreshingAgenda = true
+                                    )
+                                }
+                            }
```
steve1rm, line 79: Flag is check here and if its true we are doing a swipe to refresh so update the state and show the swipe progress indicator

presentation/build.gradle.kts
```diff
@@ -97,7 +97,7 @@ dependencies {
     implementation("io.coil-kt:coil-compose:1.4.0")
     implementation("androidx.work:work-runtime-ktx:2.8.0")
     implementation("androidx.hilt:hilt-work:1.0.0")
-
+    implementation("com.google.accompanist:accompanist-swiperefresh:0.27.0")
```
philipplackner, line 100: You don't need a dependency for that anymore, there's a normal SwipeToRefresh composable (or similar) available. Just make sure to use the latest stable compose version (I think `1.4.2` atm)

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
philipplackner: Yes that works

philipplackner (approved):

presentation/build.gradle.kts
steve1rm: @philipplackner Yes, I discovered that after I implemented it. I have the migration guide here https://google.github.io/accompanist/swiperefresh/

PR #63, opened 2023-05-06
[62] fixes issue of is in edit or save mode

steve1rm:
# Pull Request template

###  What does this PR solve?

- Displays the arrows and `save` in edit mode
- Hides the arrows and shows pencil icon in non-edit mode
- Can toggle between edit and non-edit mode.

non-edit-mode-2

edit-mode-1

### DIFF
--- a/data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
+++ b/data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
@@ -7,6 +7,7 @@ import me.androidbox.data.mapper.toUpdateEventDto
 import me.androidbox.data.remote.model.request.EventCreateRequestDto
 import me.androidbox.data.remote.model.request.EventUpdateRequestDto
 import me.androidbox.domain.agenda.model.Event
+import me.androidbox.domain.constant.UpdateModeType
 import me.androidbox.domain.work_manager.UploadEvent
 import java.util.concurrent.TimeUnit
 import javax.inject.Inject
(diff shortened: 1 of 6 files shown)


presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
```diff
@@ -294,44 +296,53 @@ class EventViewModel @Inject constructor(
...
-            startDateTime = startDateTime.toEpochSecond(),
-            endDateTime = endDateTime.toEpochSecond(),
-            remindAt = remindAt.toEpochSecond(),
-            eventCreatorId = preferenceRepository.retrieveCurrentUserOrNull()?.userId ?: "",
-            isUserEventCreator = true,
-            isGoing = true,
-            attendees = eventScreenState.value.attendees,
-            photos = eventScreenState.value.listOfPhotoUri
-        )
+        preferenceRepository.retrieveCurrentUserOrNull()?.userId?.let { userId ->
```
steve1rm, line 304: I am just wondering what to do if the userId was null. As its a nullable type is it worth checking if its null? Is it possible that it could be null. And in cases of null what is the best way to handle.
In my case I check for null and if its null show the user a snackbar or toast message saying that insert cannot proceed? Is this correct?

data/src/main/java/me/androidbox/data/worker_manager/UploadEventImp.kt
```diff
@@ -23,22 +24,25 @@ class UploadEventImp @Inject constructor(
...
-        }
-        else {
-            /** request for Create eventRequestDto */
-            val jsonAdapter = moshi.adapter(EventCreateRequestDto::class.java)
-            jsonAdapter.toJson(event.toCreateEventDto())
+        val eventRequestJson: String = when(updateModeType) {
+            UpdateModeType.CREATE -> {
+                /** request for Create eventRequestDto */
+                val jsonAdapter = moshi.adapter(EventCreateRequestDto::class.java)
+                jsonAdapter.toJson(event.toCreateEventDto())
```
philipplackner, line 33: You don't really need to create the adapter separately here if you don't reuse it

presentation/src/main/java/me/androidbox/presentation/event/viewmodel/EventViewModel.kt
philipplackner: Would this ever be the case here? If so, it would rather be a bug and not a situation which the user can really cause. Usually apps just show a generic error message in this case such as `Oops, something went wrong!`

philipplackner (approved):

PR #66, opened 2023-05-07
[64] Adds create task synchronizer

steve1rm:
# Pull Request template

###  What does this PR solve?

- Adds the synchronizer for sending created task, reminders, and events that failed to create or update, i.e offline


data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+
+        val result = if (createdTasks.await().isNotEmpty()) {
+            if(runAttemptCount > RETRY_COUNT) {
+                Result.failure()
+            }
+            else {
+                Result.retry()
+            }
```
steve1rm: If we have created tasks that need to be sync'ed. If there is a failure. Is it ok to do it in this block?

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,78 @@
...
+                    Result.success()
+                },
+                onFailure = {
+                    Result.retry()
+                }
+            )
+        }
+        else {
+            /** There was nothing to sync as the sync table was empty */
+            Result.success()
```
steve1rm: Just return success as there was no tasks to sync?

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerImp.kt
```diff
@@ -0,0 +1,46 @@
...
+import java.util.UUID
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+class TaskCreateSynchronizerImp @Inject constructor(
+    private val workManager: WorkManager
+) : TaskCreateSynchronizer {
+
+    companion object {
+        const val SYNC_INTERVAL = 30L
```
philipplackner: It makes sense to add which unit this is in, such as `SYNC_INTERVAL_MINUTES`

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,78 @@
...
+        val createdTasks: Deferred<List<TaskEntity>>
+
+        /** Get all the created tasks */
+        supervisorScope {
+            createdTasks = async {
+                taskDao.getAllTasksBySyncType(SyncAgendaType.CREATE).map {
+                    it.task
+                }
+            }
+        }
```
philipplackner: The `supervisorScope` and `async` block doesn't do anything here, since the coroutine will suspend until the tasks are retrieved anyways. You can just do `val createdTasks = taskDao.getAllTasksBySyncType(...)` as you're not executing anything in parallel here

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
philipplackner: Not really getting what this is for... Don't you want to sync these tasks here? Why return a failure if there are tasks to sync?

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
philipplackner: Yes

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,78 @@
...
+                }
+            }
+
+            responseState.fold(
+                onSuccess = {
+                    taskDao.deleteSyncTasksBySyncType(SyncAgendaType.CREATE)
+                    Result.success()
+                },
+                onFailure = {
+                    Result.retry()
```
philipplackner: If this fails, wouldn't you want to retry only deleting?

philipplackner (approved):

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
steve1rm: @philipplackner Maybe a miss understanding. I am wondering is this the best place for this:
```
            if(runAttemptCount > RETRY_COUNT) {
                Result.failure()
            }
            else {
                Result.retry()
            }
```

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
steve1rm: @philipplackner Not sure I fully understand.
Bascially, if success then delete all the created sync items from the local db, we don't need them anymore as we have already uploaded them.
However, if there is a failure, I did a `Result.retry()` to do the whole process again. Which would mean fetching from the local DB and then uploading them.

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,78 @@
...
+import kotlinx.coroutines.supervisorScope
+import me.androidbox.data.local.dao.TaskDao
+import me.androidbox.data.local.entity.TaskEntity
+import me.androidbox.data.mapper.toTaskDto
+import me.androidbox.data.remote.network.task.TaskService
+import me.androidbox.data.remote.util.CheckResult.checkResult
+import me.androidbox.data.worker_manager.constant.Constant.RETRY_COUNT
+import me.androidbox.domain.constant.SyncAgendaType
+
+@HiltWorker
```
MKiperszmid: You are going to have one worker for each agenda type?

Personally I just have 1 worker, that syncs everything

Theres not really a benefit in having 3 workers over just 1

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
MKiperszmid: Exactly what philipp mentions

Sinxe you are just making 1 call to the repository, and not multiple, ir makes no sense to have a supervisorscope and an async

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,78 @@
...
+        val result = if (createdTasks.await().isNotEmpty()) {
+            if(runAttemptCount > RETRY_COUNT) {
+                Result.failure()
+            }
+            else {
+                Result.retry()
+            }
+
+            val responseState = checkResult{
+                val createdTasksEntity = createdTasks.await()
```
MKiperszmid: Theres a big issue here

Lets say you have 10 tasks that need sync

If 8 go well, and 2 fail, you will return a failure, and still have 10 in the db as not synced. When in fact, 8 were successfuly synced

The issue will happen when 
1- Some tasks need to be created
2- The worker fails (because another failed)
3- When retrying, the creation will be done again. And it will fail since its a creation on an already created task. Meaning the worker will fail again... And again... And again. So this worker will never work

You should change the sync stuff on each item individually

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
MKiperszmid: Personally I'd add it at the start of the worker. So you can avoid calling the dao uselessly

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
MKiperszmid: Yeah if it fails, calling a retry should be fine

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
steve1rm: @MKiperszmid I have updated, Can you confirm this is correct with what I have done?

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
steve1rm: @MKiperszmid If I understand what you are saying. It will loop through and call `/create`  endpoint for each task that needs to be sync'ed. If it fails on one of them. The worker will fail and retry on all of them again? And as some of them has already been created already it will fail on the remote for those ones again. 

I am not sure I fully understand your solution. I think you might mean this:

```
        if (createdTasks.isNotEmpty()) {
            val responseState = checkResult{

                supervisorScope {
                    launch {
                        createdTasks.map { taskEntity ->
                            /* Create the task on remote */
                            taskService.createTask(taskEntity.toTaskDto())
                            /* If success will run this delete - else if loop next next one */
                            taskDao.deleteSyncTaskById(taskEntity.id)
                        }
                    }
                }
            }

            responseState.fold(
                onSuccess = {
                    /* Would this be nessecary if the items have already been deleted in the above code  */
                    taskDao.deleteSyncTasksBySyncType(SyncAgendaType.CREATE)
                    Result.success()
                },
                onFailure = {
                    Result.retry()
                }
            )
        }
```
If this is not correct can you please provide a code snippet for greater clarity?

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
philipplackner: Sorry got this wrong, I was thinking the onFailure refers to when the `deleteSyncTasksBySyncType` fails, but you're right, I was just tired yesterday 😄

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
philipplackner: Yes this looks better to me, good catch btw @MKiperszmid. Basically you need to retrieve all tasks at the beginning of the worker that need to be synced, but if some have been synced successfully before, they won't be contained in the next run of this worker if you've deleted them from the table before

philipplackner (approved):

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,150 @@
...
+import me.androidbox.data.mapper.toReminderDto
+import me.androidbox.data.mapper.toTaskDto
+import me.androidbox.data.remote.network.reminder.ReminderService
+import me.androidbox.data.remote.network.task.TaskService
+import me.androidbox.data.remote.util.CheckResult.checkResult
+import me.androidbox.data.worker_manager.constant.Constant.RETRY_COUNT
+import me.androidbox.domain.constant.SyncAgendaType
+
+@HiltWorker
+/** TODO Rename to something more readable as this is not just related to tasks */
```
MKiperszmid: Maybe you can rename it as "AgendaLocalSynchronizer"

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,150 @@
...
+    @Assisted private val workerParameters: WorkerParameters,
+    private val taskService: TaskService,
+    private val taskDao: TaskDao,
+    private val reminderService: ReminderService,
+    private val reminderDao: ReminderDao,
+) : CoroutineWorker(context, workerParameters) {
+
+    override suspend fun doWork(): Result {
+
+        /** TODO is this the correct place to have this check and to retry or fail */
```
MKiperszmid: Yes

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
MKiperszmid: Yea

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,150 @@
...
+            Result.retry()
+        }
+
+        val createdTasks: Deferred<List<TaskEntity>>
+        val updatedTasks: Deferred<List<TaskEntity>>
+        val createdReminders: Deferred<List<ReminderEntity>>
+        val updatedReminders: Deferred<List<ReminderEntity>>
+
+        /** Get all the created, updated tasks and reminders
+         *  TODO As these will run sequentially - is it better to use supervisor scope and run them in parallel? */
```
MKiperszmid: Yea that's fine, tho what I like to do is to not have "Deferred<List", and instead use actual lists and just .awaitAll()

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,150 @@
...
+
+                reminderIds.map { id ->
+                    reminderDao.getReminderById(id)
+                }
+            }
+        }
+
+        /** TODO I am using supervisor scope to launch so if one fails the other will still run
+         *  However, will the worker return failure if a single one fails even though some failed?
+         *  My checkResult will run everything in this block */
```
MKiperszmid: Yea it's fine to use a supervisorScope here

If even one fails, you should return a Result.retry()

Why? Because on the next run, since you removed some from the db, you will run only the remaining ones

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,150 @@
...
+        val responseState = checkResult {
+            supervisorScope {
+                val createTaskJob = launch {
+                    createdTasks.await().map { taskEntity ->
+                        /* Create the task on remote */
+                        taskService.createTask(taskEntity.toTaskDto())
+                        /* If success will run this delete - else if loop next next one */
+                        taskDao.deleteSyncTaskById(taskEntity.id)
```
MKiperszmid: You should only delete it, if the "createTask" was successful

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,150 @@
...
+                    }
+                }
+
+                val updateTaskJob = launch {
+                    updatedTasks.await().forEach { taskEntity ->
+                        taskService.updateTask(taskEntity.toTaskDto())
```
MKiperszmid: You should only delete it, if the "updateTask" was successful

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,150 @@
...
+                    createdReminders.await().forEach { reminderEntity ->
+                        reminderService.createReminder(reminderEntity.toReminderDto())
+                        reminderDao.deleteSyncReminderById(reminderEntity.id)
+                    }
+                }
+
+                val updateReminderJob = launch {
+                    updatedReminders.await().forEach { reminderEntity ->
+                        reminderService.updateReminder(reminderEntity.toReminderDto())
+                        reminderDao.deleteReminderById(reminderEntity.id)
```
MKiperszmid: Same as what I mentioned on Task

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,150 @@
...
+                    updatedTasks.await().forEach { taskEntity ->
+                        taskService.updateTask(taskEntity.toTaskDto())
+                        taskDao.deleteSyncTaskById(taskEntity.id)
+                    }
+                }
+
+                val createReminderJob = launch {
+                    createdReminders.await().forEach { reminderEntity ->
+                        reminderService.createReminder(reminderEntity.toReminderDto())
+                        reminderDao.deleteSyncReminderById(reminderEntity.id)
```
MKiperszmid: Same as what I mentioned on Task

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        return responseState.fold(
+            onSuccess = {
+                /** TODO Success we can delete all these from the sync table
+                 *  However, I don't think I need to do this now if
+                 *  1) I am already deleting the items individually when create/updated
+                 *  2) If the worker is success, but i.e. some reminders failed to create it will empty them
+                 *     and when the worker runs again, there will be nothing to fetch.
+                 *  */
```
MKiperszmid: You are already removing them one by one, so this should be unnecessary


" 2) If the worker is success, but i.e. some reminders failed to create it will empty them and when the worker runs again, there will be nothing to fetch."
In which case would the worker be success, but the reminder failed to create? If it fails to create, it should return a Result.retry()

MKiperszmid (approved):

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
steve1rm: @MKiperszmid I updated to perform this check for the agenda items i.e.
```
val createTaskJob = launch {
                    createdTasks.await().forEach { taskEntity ->
                        /* Create the task on remote */
                        val result = checkResult {
                            taskService.createTask(taskEntity.toTaskDto())
                        }
                        if(result.isSuccess) {
                            /* If success will run this delete - else if loop next next one */
                            taskDao.deleteSyncTaskById(taskEntity.id)
                        }
                    }
                }
```

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
steve1rm: @MKiperszmid Updated with example in the above comment.

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
steve1rm: @MKiperszmid Updated with an example in the above comment

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
steve1rm: @MKiperszmid Same as above with example

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
steve1rm: @MKiperszmid I have just add some snippet here as I want to clarify.

I have supervisor scope that is wrapped in checkResult block. Supervisor won't cancel any children. And at the end I check the responseState if it is success or failure. But if 1 launch was to fail, I am not sure how to handle that in this current implementation

```
       val responseState = checkResult {
            supervisorScope {
                val createTaskJob = launch {
                    createdTasks.await().forEach { taskEntity ->
                        /* delete if success */
                    }
                }

                val updateTaskJob = launch {
                    updatedTasks.await().forEach { taskEntity ->
                        /* delete if success */
                    }
                }

                val createReminderJob = launch {
                    createdReminders.await().forEach { reminderEntity ->
                        /* delete if success */
                    }
                }

                val updateReminderJob = launch {
                    updatedReminders.await().forEach { reminderEntity ->
                        /* delete if success */
                    }
                }

                /* do all the joins */
            }
        }

        return responseState.fold(
            onSuccess = {
                Result.success()
            },
            onFailure = {
                Result.retry()
            }
        )
    }
```

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
MKiperszmid: Perfect. That´s exactly what you need to do

data/src/main/java/me/androidbox/data/worker_manager/TaskCreateSynchronizerWorker.kt
MKiperszmid: One quick solution could be to have a boolean like:

var workerFailed = false

And if there's any fails, set it to true

And then just
```
if (workerFailed) {
    return Result.retry()
} else {
    return Result.success()
}
```

data/src/main/java/me/androidbox/data/worker_manager/AgendaSynchronizerWorker.kt
```diff
@@ -0,0 +1,166 @@
...
+
+                reminderIds.map { id ->
+                    reminderDao.getReminderById(id)
+                }
+            }
+        }
+
+        /** TODO I am using supervisor scope to launch so if one fails the other will still run
+         *  However, will the worker return failure if a single one fails even though some failed?
+         *  My checkResult will run everything in this block */
```
philipplackner: The worker only fails if you return `Result.failure()` (or if you throw an exception inside I assume)

data/src/main/java/me/androidbox/data/worker_manager/AgendaSynchronizerWorker.kt
```diff
@@ -0,0 +1,166 @@
...
+            supervisorScope {
+                val createTaskJob = launch {
+                    createdTasks.await().forEach { taskEntity ->
+                        /* Create the task on remote */
+                        val result = checkResult {
+                            taskService.createTask(taskEntity.toTaskDto())
+                        }
+                        if(result.isSuccess) {
+                            /* If success will run this delete - else if loop next next one */
+                            taskDao.deleteSyncTaskById(taskEntity.id)
```
philipplackner: Currently you don't consider the following scenario:
1. The user is in offline mode and adds a task
2. Then, while still being in offline mode, they delete the created task again
3. You'll end up with one entry for a created task and one entry with one deleted task, where both have the same task ID
4. This call will delete both, even though it only refers to deleting the created task. Therefore, the task that was actually deleted locally will not be deleted remotely since the DB entry is missing.

You can either handle this by also passing the sync type here to always only delete one entry or you already make sure to delete all other sync types of a task if it's deleted (which you might be doing, but at a different place - then don't worry)

data/src/main/java/me/androidbox/data/worker_manager/AgendaSynchronizerWorker.kt
```diff
@@ -0,0 +1,166 @@
...
+                /** TODO Success we can delete all these from the sync table
+                 *  However, I don't think I need to do this now if
+                 *  1) I am already deleting the items individually when create/updated
+                 *  2) If the worker is success, but i.e. some reminders failed to create it will empty them
+                 *     and when the worker runs again, there will be nothing to fetch.
+                 *  */
+                taskDao.deleteSyncTasksBySyncType(SyncAgendaType.CREATE)
+                taskDao.deleteSyncTasksBySyncType(SyncAgendaType.UPDATE)
+                reminderDao.deleteSyncRemindersBySyncType(SyncAgendaType.CREATE)
+                reminderDao.deleteSyncRemindersBySyncType(SyncAgendaType.UPDATE)
```
philipplackner: In which scenario would `responseState` be considered a success? If there were 0 failures? As you already mentioned, you don't need this since you already delete individual sync types if syncing succeeds. All sync types which couldn't be synced should be tried to sync with the next run of the worker

philipplackner (approved):

data/src/main/java/me/androidbox/data/worker_manager/AgendaSynchronizerWorker.kt
steve1rm: @philipplackner Inside `checkResult` if a exception is caught it will return `ResponseState.Failure(throwable)`

data/src/main/java/me/androidbox/data/worker_manager/TaskReminderSynchronizerImp.kt
```diff
@@ -0,0 +1,46 @@
...
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+class TaskReminderSynchronizerImp @Inject constructor(
+    private val workManager: WorkManager
+) : TaskReminderSynchronizer {
+
+    companion object {
+        const val SYNC_INTERVAL = 30L
+        const val TASK_CREATE_SYNC = "task_create_sync"
```
MKiperszmid, line 21: These can be private

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -243,4 +248,14 @@ class AgendaViewModel @Inject constructor(
     private fun syncFullAgendaItems() {
         fullAgendaSynchronizer.sync()
     }
+
+    private fun synAgendaTasksReminders() {
+        taskReminderSynchronizer.sync()
+    }
```
MKiperszmid, line 252: You are naming it "taskReminderSynchronizer" when in reality you sync everything. Not just tasks or reminders

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
```diff
@@ -50,6 +52,7 @@ class AgendaViewModel @Inject constructor(
         fetchAgendaItems(ZonedDateTime.now())
         syncAgendaItems()
         syncFullAgendaItems()
+        synAgendaTasksReminders()
```
MKiperszmid, line 53: Since you are calling it just once, and it's a one-liner, you don't really need a method

MKiperszmid, line 53: What's the difference between 

syncAgendaItems()
syncFullAgendaItems()
synAgendaTasksReminders()

MKiperszmid (approved):

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
steve1rm: @MKiperszmid Will refactor the naming

presentation/src/main/java/me/androidbox/presentation/agenda/viewmodel/AgendaViewModel.kt
steve1rm: @MKiperszmid 
syncAgendaItems - calls this endpoint /syncAgenda
syncFullAgendaItems - calls this endpoint /fullAgenda
syncAgendaTaskReminders (will change the name) upload all the created/updated tasks, reminders, and events. That was created offline

PR #67, opened 2023-05-09
[65]  shows snackbar with try again button

steve1rm:
# Pull Request template

###  What does this PR solve?

- Shows snack when task fails to complete
- Shows a button to allow the user to try again
- Clicking try again will try and send the task again

snackbar

### DIFF
--- a/presentation/src/main/java/me/androidbox/presentation/task/screen/TaskDetailScreen.kt
+++ b/presentation/src/main/java/me/androidbox/presentation/task/screen/TaskDetailScreen.kt
@@ -62,12 +62,10 @@ fun TaskDetailScreen(
         SnackbarHostState()
     }
     val coroutineScope = rememberCoroutineScope()
-    var snackbarJob: Job? = null
 
     LaunchedEffect(key1 = taskDetailScreenState.showSnackBar) {
         if(taskDetailScreenState.showSnackBar) {
-            snackbarJob?.cancel()
-            snackbarJob = coroutineScope.launch {
+            coroutineScope.launch {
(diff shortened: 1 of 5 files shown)


presentation/src/main/java/me/androidbox/presentation/task/screen/TaskDetailScreen.kt
```diff
@@ -54,6 +58,24 @@ fun TaskDetailScreen(
...
     val calendarStateTimeDate = rememberUseCaseState()
+    val snackbarHostState = remember {
+        SnackbarHostState()
+    }
+    val coroutineScope = rememberCoroutineScope()
+    var snackbarJob: Job? = null
+
+    LaunchedEffect(key1 = taskDetailScreenState.showSnackBar) {
+        if(taskDetailScreenState.showSnackBar) {
+            snackbarJob?.cancel()
```
steve1rm: I added a cancel here, as if the user clicks save multiple times, I just want to dismiss the current snackbar and then show it again. What do you think?

philipplackner: That's what `LaunchedEffect` already does. Every time the passed key changes, the coroutine is cancelled and relaunched

philipplackner (approved):

presentation/src/main/java/me/androidbox/presentation/task/screen/TaskDetailScreen.kt
steve1rm: @philipplackner Will remove the snackbarJob.cancel()

